PR #4, opened 2021-09-29
Cluster post config aws region updates

paigerube14:
Adding in zone a instead of d. This was not working for me on the openshift-qe aws account. Also changing to private security group. Using the public group the workload node never would become running

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -315,6 +315,18 @@ pipeline {
                 exit 1
               fi
             done
+            retries=0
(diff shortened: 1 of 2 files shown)


paigerube14:
@mffiedler are you able to review this PR as well when you get a chance? 

https://mastern-jenkins-csb-openshift-qe.apps.ocp4.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/cluster-post-config/15/console

kedark3 (approved):

PR #13, opened 2021-11-18
Taking out upgrade version parameter from write upgrade info

paigerube14:
Taking out the upgrade version parameter because it is no longer needed in my write to sheet script

Please review when you have time @kedark3


Jenkinsfile
```diff
@@ -23,6 +23,7 @@ pipeline {
         string(name: 'CI_JOB_ID', defaultValue: '', description: 'Scale-ci job id')
         string(name: 'CI_JOB_URL', defaultValue: '', description: 'Upgrade job url')
         string(name: 'LOADED_JOB_URL', defaultValue: '', description: 'Upgrade job url')
+        string(name: 'UPGRADE_VERSION', defaultValue: '', description: 'Versions cluster is upgraded to')
```
kedark3, line 26: Does user pass this param to tell us what version cluster should be upgraded to?

paigerube14: Yes, the parameter should be set by the user to see what version the cluster should've gotten to. I think I could just  tell that from doing oc commands

paigerube14:
Going to change my script to not expect that parameter because it's not really needed when I can poll the cluster

kedark3:
@paigerube14 can you please also update the PR description ?

kedark3 (approved):

PR #15, opened 2021-11-19
Adding in upgrade scripts and artifact must gather

paigerube14:
This PR is no longer going to use the svt upgrade scripts; it will move them here to use

I also added in a way to artifact the must gather if the upgrade fails 

Please take a look when you can @kedark3

### DIFF
--- a/upgrade_scripts/check_upgrade.py
+++ b/upgrade_scripts/check_upgrade.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python
+
+import time
+import yaml
+import subprocess
+import sys
+
+
+# Invokes a given command and returns the stdout
+def invoke(command):
+    try:
+        output = subprocess.check_output(command, shell=True,
+                                         universal_newlines=True)
+    except Exception as e:
+        print("Failed to run %s" % (command))
+        print("Error %s" % (str(e)))
+        return ""
+    return output
+
+
+def set_max_unavailable(max_unavailable):
+
+    merge_json ='{"spec":{"maxUnavailable":' + str(max_unavailable) + '}}'
+    invoke("oc patch machineconfigpool.machineconfiguration.openshift.io/worker --type='merge' -p='"+ merge_json + "'")
+
+
+def check_cluster_version():
+
+    output = invoke("oc get clusterversion -o yaml")
+    if output != "":
+        #parse output to get cluster version and status?
+        output = yaml.load(output, Loader=yaml.FullLoader)
+
+        for condition in output['items'][0]['status']['conditions']:
+            if "Progressing" == condition['type']:
+                print("Progressing status " + str(condition['message']))
+                break
+        for status_history in output['items'][0]['status']['history']:
+            if status_history['state'] == "Completed":
+                print('actual version ' + str(status_history['version']))
+                return status_history['version']
+            else:
+                print("State of version: " + str(status_history['state']) + " " + str(status_history['version']))
+        print("\n\n")
+        return output
+    return ""
+
+# Main function
+def check_upgrade( expected_cluster_version):
+    print("Starting upgrade check")
+    upgrade_version = check_cluster_version()
+    j = 0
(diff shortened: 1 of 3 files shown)


upgrade_scripts/check_upgrade.py
```diff
@@ -0,0 +1,117 @@
...
+        print("Failed to run %s" % (command))
+        print("Error %s" % (str(e)))
+        return ""
+    return output
+
+
+def set_max_unavailable(max_unavailable):
+
+    merge_json ='{"spec":{"maxUnavailable":' + str(max_unavailable) + '}}'
+    invoke("oc patch machineconfigpool.machineconfiguration.openshift.io/worker --type='merge' -p='"+ merge_json + "'")
```
kedark3, line 24: Do we not want to check the return status of the `invoke` ? I feel like `invoke` should return 0 or non-zero return code and we should check that upon return. Maybe a 2 tuple of `(output, rc)`

upgrade_scripts/check_upgrade.py
```diff
@@ -0,0 +1,117 @@
...
+
+def check_cluster_version():
+
+    output = invoke("oc get clusterversion -o yaml")
+    if output != "":
+        #parse output to get cluster version and status?
+        output = yaml.load(output, Loader=yaml.FullLoader)
+
+        for condition in output['items'][0]['status']['conditions']:
+            if "Progressing" == condition['type']:
```
kedark3, line 35: How about rewriting this as : 
```             if condition['type'] == "Progressing" : ```

More readable.

File: upgrade_scripts/check_upgrade.py [hunk 3]
```diff
@@ -0,0 +1,117 @@
...
+                print("Progressing status " + str(condition['message']))
+                break
+        for status_history in output['items'][0]['status']['history']:
+            if status_history['state'] == "Completed":
+                print('actual version ' + str(status_history['version']))
```
kedark3, line 40: Also how about using `f` strings over `+` for print statements/strings everywhere?

upgrade_scripts/check_upgrade.py
```diff
@@ -0,0 +1,117 @@
...
+        upgrade_version = check_cluster_version()
+        time.sleep(30)
+        j += 1
+    return 1
+
+
+def wait_for_co_ready():
+
+    counter = 0
+    wait_num = 30
```
kedark3, line 68: `wait_num = 30` this could move into the function signature as is, so it will still take default 30 but now it makes it configurable. Same below as well.

upgrade_scripts/check_upgrade.py
```diff
@@ -0,0 +1,117 @@
...
+            return
+        print("Waiting 30 seconds for nodes to become ready and scheduling enabled")
+        time.sleep(30)
+        counter += 1
+    print("ERROR: Nodes were still not ready and scheduling enabled after 30 minutes")
+
+def wait_for_replicas(machine_replicas, machine_name):
+    counter = 0
+    wait_num = 60
+    replicas = invoke("oc get " + machine_name + " -n openshift-machine-api -o jsonpath={.status.availableReplicas}")
```
kedark3, line 96: f strings here as well.

kedark3 (commented):
Great work Paige, leaving some comments. Please review.

paigerube14:
@kedark3 added in all your comments, please take a look when you get a chance. Will work on other PR comments

paigerube14:
Should all use f string now, PTAL @kedark3 

On line 21 I couldnt use f string because of the below error

```

(venv3) prubenda@prubenda1-mac ocp-qe-perfscale-ci % python upgrade_scripts/check_upgrade.py
  File "upgrade_scripts/check_upgrade.py", line 23
    merge_json = f'{"spec":{"maxUnavailable": {{maxUnavailable}} }}'
                 ^
SyntaxError: f-string: expressions nested too deeply

```

kedark3 (approved):

PR #31, opened 2021-12-17
Adding correct storage class for monitoring

paigerube14:
Adding in correct storage class for azure when moving things to the infra nodes

The managed premium is on the cluster by default so using that type


skordas (approved):
/lgtm

PR #32, opened 2021-12-17
Adding correct storage class for azure

paigerube14:
Need to pass the correct parameters to the post config script with the default storage class for azure

This PR goes along with https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/31


skordas (approved):
/lgtm

PR #49, opened 2022-01-13
Post config azure fix with storage

paigerube14:
Going back to Premium_LRS storage type, need storageaccounttype instead of storageclass type. Other PR for in cluster worker scale coming as well


skordas (approved):
/lgtm

PR #51, opened 2022-01-18
Adding worker and master node counts for creating flexy 

paigerube14:
Adding in master and worker counts to flexy installation. This is helpful for things such as 5 master clusters


skordas (approved):
/lgtm

PR #62, opened 2022-02-07
Adding ovn type for alibaba

paigerube14:
Currently flexy doesn't support any other alibaba installer type other than fips-ovn-ci, so adding in this specific type for alibaba


skordas (approved):
/lgtm

PR #64, opened 2022-02-09
Adding env vars passing to different jobs

paigerube14:
For vsphere we need to export certain proxy's. To be able to call the worker scaler on the cluster from within the node density test we need to pass the env-vars set from node-density


skordas (approved):
/lgtm

PR #80, opened 2022-02-16
Adding test case and parameter suggestions

paigerube14:
Adding polarion test case link and parameter sugguestions


qiliRedHat:
@paigerube14 Hi Paige, thanks for updating the Job with the parameter suggestion, it's very helpful.

PR #88, opened 2022-02-23
Adding env vars to write to sheet 

paigerube14:
Vsphere upgrade fails when running write to sheet. Need to be able to pass env_vars from upgrade to write to sheet


skordas (approved):
/lgtm

PR #91, opened 2022-03-03
Loaded upgrades with Ash 

paigerube14:
Adding in Ash flexy installation as well as some formatting to separate the parameters by type. See screenshot below of new look 

..../scale-ci/job/paige-e2e-multibranch/job/loaded-upgrade/285/console (failure was due to storage-perf being taken out, flexy installation and passing of env variables passed)

Screen Shot 2022-03-02 at 4 27 55 PM (3)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,10 +19,21 @@ pipeline{
     agent any
 
     parameters {
+        separator(name: "PRE_BUILT_FLEXY_ENV", sectionHeader: "Pre Built Flexy", sectionHeaderStyle: """
+				font-size: 18px;
+				font-weight: bold;
+				font-family: 'Orienta', sans-serif;
+			""")
         string(name: 'BUILD_NUMBER', defaultValue: '', description: 'Build number of job that has installed the cluster.')
+
+        separator(name: "BUILD_FLEXY", sectionHeader: "Build Flexy Parameters", sectionHeaderStyle: """
+				font-size: 18px;
+				font-weight: bold;
+				font-family: 'Orienta', sans-serif;
+			""")
         string(name: 'OCP_PREFIX', defaultValue: '', description: 'Name of ocp cluster you want to build')
         string(name: 'OCP_VERSION', defaultValue: '', description: 'Build version to install the cluster.')
-        choice(choices: ['','aws', 'azure', 'gcp', 'osp', 'alicloud', 'ibmcloud', 'vsphere'], name: 'CLOUD_TYPE', description: '''Cloud type (As seen on https://gitlab.cee.redhat.com/aosqe/flexy-templates/-/tree/master/functionality-testing/aos-4_9, after ""-on-") <br/>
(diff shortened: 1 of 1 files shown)


qiliRedHat:
@paigerube14 Thank you for adding ASH! And the new look is quite organized!
Some comments:
1. For Scale-CI Job Options, CI_TYPE I remember you mentioned there is some reason we can't make it a drop down list, can we added all the ci types in the description? It can save time(I'm sometimes not confident about my memory and need to check the names otherwhere) and avoid typo(can copy and paste).
2. For Scale-CI Job Options, JOB_ITERATIONS NODE_COUNT PODS_PER_NODE, can we add in description this is for which CI_TYPE and the recommendations? For example for JOB_ITERATIONS, 4*worker node number.
3. ENV_VARS: It can include the parameters for the scale-ci builds so I think it can be put under Scale-CI Job Options, but it can also include proxy settings for all the onward jobs. I'm not quite sure where is better to put it. Maybe keep it as now :)

The above comments are not hurry, you can consider them in later PR.  For the change in this PR, it looks good to me. Let me know if you want to merge this PR now or later. I'll approve. Really appreciate your effort to continuously and quickly enhance our automation!

qiliRedHat:
Sorry it got merged. Curious about why bot added the approved label without /approve. Found the answer that "An approver's /lgtm is simultaneously interpreted as an /approve" in https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#the-code-review-process.

paigerube14:
> @paigerube14 Thank you for adding ASH! And the new look is quite organized! Some comments:
> 
> 1. For Scale-CI Job Options, CI_TYPE I remember you mentioned there is some reason we can't make it a drop down list, can we added all the ci types in the description? It can save time(I'm sometimes not confident about my memory and need to check the names otherwhere) and avoid typo(can copy and paste).
> 2. For Scale-CI Job Options, JOB_ITERATIONS NODE_COUNT PODS_PER_NODE, can we add in description this is for which CI_TYPE and the recommendations? For example for JOB_ITERATIONS, 4*worker node number.
> 3. ENV_VARS: It can include the parameters for the scale-ci builds so I think it can be put under Scale-CI Job Options, but it can also include proxy settings for all the onward jobs. I'm not quite sure where is better to put it. Maybe keep it as now :)
> 
> The above comments are not hurry, you can consider them in later PR. For the change in this PR, it looks good to me. Let me know if you want to merge this PR now or later. I'll approve. Really appreciate your effort to continuously and quickly enhance our automation!

Hey @qiliRedHat, see my responses to each question below

1.  The CI_TYPE should be a drop down list (see screenshot below). The drop downs don't show up when you do rebuild because it takes the string version of the parameter but if you click on build with parameters from the beginning you'll be able to see the options (hopefully) 

Screen Shot 2022-03-07 at 9 41 27 AM (2)

2. Yes definitely think we should add sugguestions on parameters. I think the best place to do it is in loaded-upgrade.  Adding the details to each of the scale-ci jobs is time consuming but will also keep working on that

3. I think I should move up the scale-up and scale-down parameters above the scale-ci because that's the order it runs it. What do you think. I think the ENV_VARS and JENKINS_VARIABLES might be better kept at the end to apply to all jobs? What do you think

qiliRedHat:
@paigerube14 Regarding discussion in above comment.
1. Thanks:) I usually do a rebuild and didn't notice the difference.
2. Agree with you.
3. Sounds reasonable. Let's do it like you said.

PR #100, opened 2022-03-15
New Kube burner job with parameterized workload variable

paigerube14:
This will replace the pod-density, cluster-density, node-density, node-density-heavy, max-namespaces and max-services branches into one common location

This will allow us to more easily edit all these workloads in one common place. 

See passing job here: Jenkins/job/scale-ci/job/paige-e2e-multibranch/job/kube-burner-2/5/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,149 @@
+@Library('flexy') _
+
+// rename build
+def userId = currentBuild.rawBuild.getCause(hudson.model.Cause$UserIdCause)?.userId
+if (userId) {
+  currentBuild.displayName = userId
+}
+
+def RETURNSTATUS = "default"
+
+pipeline {
+  agent none
+
+  parameters {
+        string(name: 'BUILD_NUMBER', defaultValue: '', description: 'Build number of job that has installed the cluster.')
+        choice(choices: ["cluster-density","node-density","node-density-heavy","pod-density","pod-density-heavy","max-namespaces","max-services"], name: 'WORKLOAD', description: '''Type of kube-burner job to run''')
+        booleanParam(name: 'WRITE_TO_FILE', defaultValue: false, description: 'Value to write to google sheet (will run https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/write-scale-ci-results)')
+
+        string(name:'JENKINS_AGENT_LABEL',defaultValue:'oc45',description:
+        '''
+        scale-ci-static: for static agent that is specific to scale-ci, useful when the jenkins dynamic agen
+ isn't stable<br>
+        4.y: oc4y || mac-installer || rhel8-installer-4y <br/>
+            e.g, for 4.8, use oc48 || mac-installer || rhel8-installer-48 <br/>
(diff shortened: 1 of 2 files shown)


qiliRedHat:
@paigerube14 I have some comments
1. Shall we change ITERATIONS to some other word, like VARIABLE or some else, you're better on naming than me:). The reason is I think it is a little confusing. Because for cluster-density, it really does it in ‘iteration’ ways, but for others, they are not. 

2. In the description, I feel 'iteration' is confusing. Can we do it like below?  If you think the last sentence of each description makes it too long, pls remove it, anyway they are in the link below.

This variable configures parameter needed for each type of workload. By default 1000.
pod-density: Set variable PODS. Set to 200 * num_workers, work up to 250 * num_workers including existing pods on a worker. Creates as many "sleep" pods as configured in this environment variable.
cluster-density: Set variable JOB_ITERATIONS. Set to 4 * num_workers. This variable configures the number of cluster-density jobs iterations to perform (1 namespace per iteration). 
max-namespaces: Set variable NAMESPACE_COUNT. Set to ~30 * num_workers. The number of namespaces created by Kube-burner is defined by this variable.
max-services: Set variable SERVICE_COUNT. Set to 200 * num_workers, work up to 250 * num_workers including existing pods on a worker. Creates n-replicas of an application deployment (hello-openshift) and a service in a single namespace.
Read here for detail of each variable:
https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/kube-burner/README.md

paigerube14:
@qiliRedHat 

1. I agree that we should maybe make the parameter for general so it doesn't get confusing with cluster-density vs other tests
2. I like the second part of each job description you added. Might try to condense a little so it's not as long, just not sure how much people will be willing to read 

Will work on that today. Thanks for the points

qiliRedHat:
@paigerube14 I just noticed the` NODE_COUNT` is now the worker nodes numbers in the cluster.  Previously in run_nodedensity-heavy_test_fromgit.sh, it is hardcoded to 4, so that we must pass this parameters.
So I think we don't need to have this it on the jenkins UI now, so we don't need the separate `Node Density Job Options` section. We can pass it with ENV_VARS if needed in some rare cases.
`    NODE_COUNT=${NODE_COUNT:-$(kubectl get node -l node-role.kubernetes.io/worker,node-role.kubernetes.io/infra!=,node-role.kubernetes.io/workload!= -o name | wc -l)}`
https://github.com/cloud-bulldozer/e2e-benchmarking/blob/544ebf9a0a47f5fbf704c346cb93e319f9053281/workloads/kube-burner/run.sh#L14-L26

mffiedler:
> @paigerube14 I just noticed the` NODE_COUNT` is now the worker nodes numbers in the cluster. Previously in run_nodedensity-heavy_test_fromgit.sh, it is hardcoded to 4, so that we must pass this parameters. So I think we don't need to have this it on the jenkins UI now, so we don't need the separate `Node Density Job Options` section. We can pass it with ENV_VARS if needed in some rare cases.

I suggest we leave it in for now and see if it is useful.   I personally use NODE_COUNT quite a bit when working on non-standard sized clusters.   But, that's just me :-).  

@qiliRedHat or @paigerube14 If we are done you can cancel the hold and the PR should merge.

PR #103, opened 2022-03-16
Scale jobs in loaded-upgrade

paigerube14:
This is adding in calls to scale up and the 2 more new general scale-ci jobs (kube-burner and network-perf). Those 2 pr's need to go in before this one and I'll update the build links 

For etcd-perf and router-perf I took out the passing of scale-up and scale-down so it doesn't scale twice 

I wasn't sure where scale-down fits into this workload. Any ideas? 

NOTE: Wanting to get this initial PR in before adding in a cerberus check as well as getting a must-gather if any failures

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,14 +1,20 @@
 
 def aws = null
 def install = null
+def scale_up = null
 def loaded_ci = null
+def health_check = null
 def upgrade_ci = null
 def destroy_ci = null
+def must_gather = null
 def build_string = "DEFAULT"
 def load_result = "SUCCESS"
 def loaded_url = ""
 def upgrade_url = ""
+def must_gather_url = ""
 def proxy_settings = ""
+def status = "PASS"
+
 
 def userId = currentBuild.rawBuild.getCause(hudson.model.Cause$UserIdCause)?.userId
 if (userId) {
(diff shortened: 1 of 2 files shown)


mffiedler:
Scale up/down is meaningless for the etcd-perf test.
For router-perf it is potentially useful as the large tests do scale the cluster up to 30+ or 50+ nodes

paigerube14:
Gotcha, I will completely take out of etcd-perf for sure. I was thinking that if you wanted to scale and run a benchmark (ie router-perf) you would run it through this job so this job would do the scaling and run router-perf instead of running router-perf and the scaling being done in router-perf specifically. Not sure if that makes sense

Thinking about it more, should we just take out the passing of the scale-up and scale-down parameter from this job but leave them as parameters in the scale-ci jobs themselves. So that you aren't forced to run through the loaded-upgrade job if you wanted to scale your cluster before running the benchmark?

qiliRedHat:
@paigerube14 I prefer 'Thinking about it more, should we just take out the passing of the scale-up and scale-down parameter from this job but leave them as parameters in the scale-ci jobs themselves. ', it's more flexible. Now you merged the kube-burner branches, there is less effort maintaining that in each branch. 

For scale-down, I never used it, I always destroy directly after manual post check.

paigerube14:
@qiliRedHat I think we should still add it to this job to make it easier to track if there is a failure from scaling the cluster. I had added it back into kube-burner to make it flexible (like you sugguested)  to be able to run scale and any kube-burner job easily as well. But adding here would also be flexible and it is not passing the number workers to scale, so we don't double scale. Hope this makes sense and we can re-evaluate in the future if this no longer makes sense

PR #104, opened 2022-03-17
Adding scale up and down back in

paigerube14:
Based on Qiujie's comment [here](https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/103#issuecomment-1070509912), adding back the scaleup and down into this job to make running more flexible. 

Just won't pass the parameters from the loaded-upgrade so we don't run into double scaling isuses

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,6 +54,11 @@ pipeline {
                SOMEVARn='envn-test'<br>
                </p>'''
             )
+        string(name: 'SCALE_UP', defaultValue: '0', description: 'If value is set to anything greater than 0, cluster will be scaled up before executing the workload.')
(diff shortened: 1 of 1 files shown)


skordas (approved):
/lgtm

qiliRedHat:
@paigerube14 Thanks for adding them back. Code looks good but why I didn't see this in the Jenkins sandbox https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/kube-burner-2/?

paigerube14:
When you make updates to the parameters you have to run a build  for Jenkins to see the parameters have updated and then they would be at the build with parameters the next time. I usually just give blank build number so it doesn't actually run the test at all

PR #114, opened 2022-04-07
Adding in a backup way to get UUID (helps with concurrent builds)

paigerube14:
Adding in a couple of updates for failure cases of not finding UUID for the run. This will also help with concurrent builds when the benchmark is cleaned up after every application run we will still be able to get the pod_ready informaiton


skordas (approved):
/lgtm

PR #115, opened 2022-04-08
Adding correct result to not get null error

paigerube14:
Fixing incorrect variable if scale-ci job fails; found by Nathan so adding him to list of reviewers

@mffiedler do I need to also add him in the PROW owners list or is just here ok?


mffiedler:
Just here should be ok for now.  @nathan-weinberg you need to request membership to the openshift-qe org.

nathan-weinberg (approved):
Looks like my review permissions are working :+1:

PR #120, opened 2022-04-18
Small fix for router perf getting UUID

paigerube14:
There are some cases where UUID might be capitalized in the json we get to get uuid for router perf. Adding in return of lower case uuid, upper case and adding in error case option 

Example of failure: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/write-scale-ci-results/830/console


skordas (approved):
/lgtm

PR #123, opened 2022-04-21
Adding getting pod averages in error cases

paigerube14:
Currently when the kube-burner tests finish it cleans up the benchmark-operator namespace and all objects that I was using to get the pod ready latencies. I have added in back up ways to get the UUID from the output but now need to pass it to be able to get the pod latencies of the job properly

I created a PR to only cleanup the namespace if you want to cleanup all the objects as well
https://github.com/cloud-bulldozer/e2e-benchmarking/pull/381


paigerube14:
This also fixes not writing proper results for etcd-perf

skordas (approved):
/lgtm

PR #133, opened 2022-05-03
Upgrade EUS clusterversion check

paigerube14:
The check to validate the clusterversion is in the list of possible upgrades for the channel type was not working properly. This fix better checks the value returned from _oc adm upgrade_


skordas (approved):
/lgtm

PR #135, opened 2022-05-03
Initial commit of NOPE tool

nathan-weinberg:
This PR:
- Creates an initial skeleton of the NOPE tool
- Adds a `requirements.txt` file for the NOPE tool dependencies
- Updates `README.md` with details about the NOPE tool
- Updates `.gitignore` to prevent performance data from being pushed to GitHub
- Resolves https://issues.redhat.com/browse/NETOBSERV-327


scripts/nope.py
```diff
@@ -0,0 +1,104 @@
...
+	parser = argparse.ArgumentParser(description='Network Observability Prometheus and Elasticsearch tool (NOPE)')
+
+	# set customization flags
+	parser.add_argument("--url", type=str, help='Cluster URL', required=True)
+	parser.add_argument("--token", type=str, help='Token for Bearer auth', required=True)
+
+	# parse arguments
+	args = parser.parse_args()
+	CLUSTER_URL = args.url
+	TOKEN = args.token
```
paigerube14, line 98: might be nice here to set a default value by getting response of _oc whoami -t_ but if the user does pass the -t option to overwrite default. Same with CLUSTER_URL. Thoughts?

scripts/nope_requirements.txt
```diff
@@ -0,0 +1,2 @@
+argparse==1.4.0
```
memodi, line 1: change filename to be requirements.txt

memodi (changes requested):

scripts/nope.py
```diff
@@ -0,0 +1,104 @@
...
+import sys
+import json
+import pathlib
+import argparse
+import requests
+
+
+CLUSTER_URL = ''
+TOKEN = ''
+CPU_QUERY = 'pod:container_cpu_usage:sum{namespace="network-observability"}'
```
paigerube14, line 12: maybe we can add namespace as a parameter we pass in so this script can be more versatile and be used in any namespace.

scripts/nope.py
```diff
@@ -0,0 +1,104 @@
...
+MEM_QUERY = 'sum(container_memory_working_set_bytes{namespace="network-observability",container="",}) BY (pod)'
+LOKI_QUERY = 'kubelet_volume_stats_used_bytes{namespace="network-observability", persistentvolumeclaim="loki-store"}'
```
paigerube14, line 14: I'm not exactly sure about how NO is set up, but maybe we should add in a check to see if the loki-store pvc is on the cluster before running this query? Trying now on my cluster that that's not set up and seeing how it would react

scripts/nope.py
```diff
@@ -0,0 +1,104 @@
...
+
+def run_query(query_type=''):
+
+	# intialize request vars
+	headers = {"Authorization": f"Bearer {TOKEN}"}
+	params = {}
+	endpoint = f"https://prometheus-k8s-openshift-monitoring.apps.{CLUSTER_URL}/api/v1/query"
+
+	# build payload based on query type
+	if query_type == "cpu":
```
paigerube14, line 25: Thoughts on making this set up a 2d list or dictionary that we could loop through. 

Ex.) Dictionary set up 
query_list = { {name: cpu, query: CPU_QUERY},  {name: mem, query: MEM_QUERY}, {name: loki, query: LOKI_QUERY}}

Not sure if you really need the name key value here. 

```
for ql in query_list: 
  params = {'query': ql['query']}
  requests... 
```

scripts/nope.py
```diff
@@ -0,0 +1,104 @@
...
+
+if __name__ == '__main__':
+
+	# initialize argument parser
+	parser = argparse.ArgumentParser(description='Network Observability Prometheus and Elasticsearch tool (NOPE)')
+
+	# set customization flags
+	parser.add_argument("--url", type=str, help='Cluster URL', required=True)
+	parser.add_argument("--token", type=str, help='Token for Bearer auth', required=True)
+
```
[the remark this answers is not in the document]
paigerube14: Mehul copied my thoughts exactly. But definitely think we need start and end time here as well!

scripts/nope.py
memodi: yeah, it's something user has to deploy manually as PVC kind.  
For this tool to be versatile, we'll probably need to develop more or less a framework with some libraries/classes where user could add define all the metrics and methods to be collected and the script (for e.g.: netobserv.py) will simply call appropriate class based on argument and desired class.

scripts/nope.py
memodi: that a great idea @paigerube14 , see my comment here https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/135#discussion_r865103368 , let me know what you and @nathan-weinberg think.

scripts/nope.py
nathan-weinberg: @paigerube14 I like the idea, but I think it might be over-engineering for right now

scripts/README.md
```diff
@@ -4,4 +4,12 @@ The purpose of scripts in this directory is to measure [network-observability](h
...
-3. router-perf
\ No newline at end of file
+3. router-perf
+
+## Network Observability Prometheus and Elasticsearch tool (NOPE)
+The Network Observability Prometheus and Elasticsearch tool, or NOPE, is a Python program that is used for collecting and sharing performance data for a given OpenShift cluster running the Network Observability Operator, using Prometheus queries for collection and Elasticsearch servers for sharing. Raw JSON files are written to the `data/` directory in the project - note this directory will be created automatically if it does not already exist.
+
+### Running the tool
+1. Ensure you have Python 3.9+ and Pip installed (verify with `python --version` and `pip --version`)
+2. Install requirements with `pip install -r scripts/nope_requirements.txt`
```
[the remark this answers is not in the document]
nathan-weinberg: Yes will do

scripts/nope.py
nathan-weinberg: Similar to https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/135#discussion_r865051600 - let's discuss there

scripts/nope.py
paigerube14: yep exactly what I was thinking, make network-observability default but add optional argument to pass other namespaces

scripts/nope.py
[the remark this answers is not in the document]
nathan-weinberg: @paigerube14 @memodi So basically this is a scoping question - do we want assume users have set `kubeconfig` and logged into the cluster as `kubeadmin` **prior** to running the tool? If we make this a requirement, we can programmatically extract both of these arguments - I do like the idea. I would like to implement this as a followup if that is alright with you both, so we can get the tool merged and begin using it. I reckon I could get this done later this week or next. Let me know your thoughts in this thread, in the meantime I will push up other suggested changes.

scripts/nope.py
memodi: even for network-observability, there's nothing stopping for being deployed in different NS other than `network-observability` , so +1 for making NS configurable. thanks!

nathan-weinberg: I will update the PR to do this @paigerube14 @memodi - thanks!

scripts/nope.py
memodi: I am fine changing at later stage, it would be lot of refactoring though because it would be more or less a design change.

scripts/nope.py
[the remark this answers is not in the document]
memodi: Yes, it's fine with that assumption for now and already noted in README, so we should be good for this PR and come back later.

scripts/README.md
```diff
@@ -4,4 +4,12 @@ The purpose of scripts in this directory is to measure [network-observability](h
...
 2. node-density-heavy (using kube-burner)
-3. router-perf
\ No newline at end of file
+3. router-perf
+
+## Network Observability Prometheus and Elasticsearch tool (NOPE)
+The Network Observability Prometheus and Elasticsearch tool, or NOPE, is a Python program that is used for collecting and sharing performance data for a given OpenShift cluster running the Network Observability Operator, using Prometheus queries for collection and Elasticsearch servers for sharing. Raw JSON files are written to the `data/` directory in the project - note this directory will be created automatically if it does not already exist.
+
+### Running the tool
+1. Ensure you have Python 3.9+ and Pip installed (verify with `python --version` and `pip --version`)
```
paigerube14, line 13: is 3.9 necessary for this tool? Think in the jenkinsfile we specifically create python3.8.12 virtualenv.  I know we have hit a lot of issues with not being able to change the specific python version on Jenkins. That's why we do this: https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/f1199c911417a946c52197a560b527d4203cdb17/Jenkinsfile#L157

nathan-weinberg: Yes, there are 3.9 specific functions being used here - we'll have to change the above line, thanks for pointing this out. @memodi ^^

memodi: @nathan-weinberg - could you point out what is python 3.9 specific in this PR?

nathan-weinberg: https://docs.python.org/3/library/stdtypes.html#str.startswith

nathan-weinberg: was introduced in 3.9

scripts/README.md
```diff
@@ -4,4 +4,12 @@ The purpose of scripts in this directory is to measure [network-observability](h
...
+2. Install requirements with `pip install -r scripts/requirements.txt`
+3. Run the tool with `./scripts/nope.py` with the `--url` arg being your cluster's URL and the `--token` arg being your token for Bearer authentication. You can get the token by logging into your cluster as `kubeadmin` and running `oc whoami -t`.
```
paigerube14, line 15: maybe adding a suggestion on how to get cluster_url similar to what you did for the token in this description

scripts/README.md
memodi: okay, sounds good. I'll update Jenkinsfile to 3.9.12

PR #136, opened 2022-05-04
Added ability for NOPE to autoset cluster URL and token

nathan-weinberg:
Resolves NETOBSERV-331

### DIFF
--- a/scripts/nope.py
+++ b/scripts/nope.py
@@ -3,6 +3,7 @@
 import sys
 import json
 import pathlib
+import subprocess
 import argparse
 import requests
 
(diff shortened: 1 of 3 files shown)


scripts/nope.py
```diff
@@ -91,19 +92,33 @@ def main():
...
-	parser.add_argument("--token", type=str, help='Token for Bearer auth', required=True)
+	parser.add_argument("--url", type=str, default='', help='Override for Cluster URL', required=False)
+	parser.add_argument("--token", type=str, default='', help='Override for Bearer auth token', required=False)
 	parser.add_argument("--namespace", type=str, default='network-observability', help='Namespace for query - defaults to "network-observability"', required=False)
 
 	# parse arguments
 	args = parser.parse_args()
+
 	CLUSTER_URL = args.url
+	if CLUSTER_URL == '':
```
memodi, line 103: IMO, `if CLUSTER_URL is None` is much cleaner to check for empty strings.

scripts/nope.py
```diff
@@ -91,19 +92,33 @@ def main():
 	parser = argparse.ArgumentParser(description='Network Observability Prometheus and Elasticsearch tool (NOPE)')
 
 	# set customization flags
-	parser.add_argument("--url", type=str, help='Cluster URL', required=True)
-	parser.add_argument("--token", type=str, help='Token for Bearer auth', required=True)
+	parser.add_argument("--url", type=str, default='', help='Override for Cluster URL', required=False)
+	parser.add_argument("--token", type=str, default='', help='Override for Bearer auth token', required=False)
```
memodi, line 96: same comment as above.

memodi (changes requested):

scripts/nope.py
```diff
@@ -91,19 +92,33 @@ def main():
...
 	# parse arguments
 	args = parser.parse_args()
+
 	CLUSTER_URL = args.url
+	if CLUSTER_URL is None:
+		CLUSTER_URL = subprocess.run(['oc', 'get', 'infrastructure', 'cluster', '-o', 'jsonpath="{.status.apiServerURL}"'], capture_output=True, text=True).stdout
+		print(f"CLUSTER_URL (raw): {CLUSTER_URL}")
+
 	TOKEN = args.token
+	if TOKEN is None:
```
paigerube14, line 108: Could also do `if not TOKEN:` but either works...

nathan-weinberg: Yeah functionally they are the same - we could change it - do you think we should?

paigerube14: nah think it's fine!

PR #139, opened 2022-05-05
Adding getting pod averages in error cases by time in job output

paigerube14:
Updating getting uuid and pod latencies timing by using creation time in the job output parameter passed. 

Fixes writing out for network-perf and concurrent-builds


skordas (approved):
Let's try it :)
/lgtm

PR #140, opened 2022-05-05
Get Prometheus queries for NOPE from YAML file

nathan-weinberg:
Resolves https://issues.redhat.com/browse/NETOBSERV-335


scripts/nope.py
```diff
@@ -2,87 +2,58 @@
...
-
-
 def main():
 
 	# get prometheus data
-	cpu_data = get_cpu_usage()
-	mem_data = get_mem_usage()
-	loki_data = get_loki_store()
+	num_queries = range(len(QUERIES))
+	for index in num_queries:
```
memodi, line 43: nit: this is fine, but I don't think you're using index, you can simply do: `for query in QUERIES`, more simple IMO

memodi (approved):

PR #116, opened 2022-04-12
Propagate false for write to file

paigerube14:
Adding propogate false for write to file job to not have write to file affect final result of kube-burner job

PR #142, opened 2022-05-09
Propagate false for write to file - router-perf

qiliRedHat:
Adding propogate false for write to file job to not have write to file affect final result of router-perf job.
Similar to https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/116


qiliRedHat:
@paigerube14 PTAL

PR #158, opened 2022-05-18
Add range query ability to NOPE

nathan-weinberg:
Few changes here:
- Disabled SSL warnings
- Added debug mode
- Changed CLUSTER_URL to THANOS_URL
- Minor changes to TOKEN autodetection and variable cleaning
- Added ability to execute range queries


scripts/nope.py
```diff
@@ -62,35 +84,58 @@ def main():
 	parser = argparse.ArgumentParser(description='Network Observability Prometheus and Elasticsearch tool (NOPE)')
 
 	# set customization flags
-	parser.add_argument("--url", type=str, help='Override for Cluster URL', required=False)
-	parser.add_argument("--token", type=str, help='Override for Bearer auth token', required=False)
-	parser.add_argument("--yaml_file", type=str, default='netobserv-metrics.yaml', help='YAML file from which to source Prometheus queries - defaults to "netobserv-metrics.yaml"', required=False)
+	parser.add_argument("--url", type=str, help='Override for Thanos URL')
```
memodi, line 87: I wonder, if it makes sense to remove `--url` and `--token` since you have added both ways to collect token depending upon user/non-user workloads. also, thanos can be standard URL.

memodi (changes requested):

scripts/nope.py
nathan-weinberg: @memodi maybe we can keep it just in case for overrides? but i'm fine either way

PR #168, opened 2022-05-27
Added netobserv operator data to JSON payload

nathan-weinberg:
Resolves [NETOBSERV-375](https://issues.redhat.com/browse/NETOBSERV-375)

### DIFF
--- a/scripts/nope.py
+++ b/scripts/nope.py
@@ -59,7 +59,30 @@ def run_query(query):
 	return data.json()
 
 
-def get_operator_info():
+def run_commands(commands, outputs={}):
+
+	# iterate through commands dictionary
+	for command in commands:
+		if DEBUG:
+			print(f"\nExecuting command '{' '.join(commands[command])}' to get {command} data")
+		result = subprocess.run(commands[command], capture_output=True, text=True)
+
+		# record command stdout if execution was succesful
+		if result.returncode == 0:
+			output = result.stdout[1:-1]
+			if DEBUG:
+				print(f"Got back result: {output}")
+			outputs[command] = output
+
+		# otherwise raise an Exception with stderr
+		else:
+			raise Exception(f"Command '{command}' execution resulted in stderr output: {result.stderr}")
+
+	# if all commands were successful return outputs dictionary
+	return outputs
+
+
+def get_netobserv_env_info():
 
 	# intialize info and base_commands objects
 	info = {}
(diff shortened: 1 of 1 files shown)


nathan-weinberg:
Example of data being generated:
```
    "netobservEnv": {
        "release": "main",
        "flp_kind": "DaemonSet",
        "loki_pvc_cap": "1Gi",
        "agent": "ipfix",
        "sampling": "400",
        "cache_active_time": "60s",
        "cache_max_flows": "100"
    }
```

scripts/nope.py
```diff
@@ -59,8 +59,46 @@ def run_query(query):
...
+	base_commands = {
+		"release": ['oc', 'get', 'pods', '-l', 'app=network-observability-operator', '-o', 'jsonpath="{.items[0].metadata.labels.version}"'],
+		"flp_kind": ['oc', 'get', 'flowcollector', '-o', 'jsonpath="{.items[*].spec.flowlogsPipeline.kind}"'],
+		"loki_pvc_cap": ['oc', 'get', 'pvc/loki-store', '-o', 'jsonpath="{.status.capacity.storage}"'],
+		"agent": ['oc', 'get', 'flowcollector', '-o', 'jsonpath="{.items[*].spec.agent}"'],
+	}
+
+	# collect data from cluster about netobserv operator
+	for command in base_commands:
+		info[command] = subprocess.run(base_commands[command], capture_output=True, text=True).stdout[1:-1]
```
memodi, line 75: I would suggest to check stderr as well in case of errors before checking for stdout.

scripts/nope.py
```diff
@@ -59,8 +59,46 @@ def run_query(query):
...
+	agent = info["agent"]
+	agent_commands = {
+		"sampling": ['oc', 'get', 'flowcollector', '-o', f'jsonpath="{{.items[*].spec.{agent}.sampling}}"'],
+		"cache_active_time": ['oc', 'get', 'flowcollector', '-o', f'jsonpath="{{.items[*].spec.{agent}.cacheActiveTimeout}}"'],
+		"cache_max_flows": ['oc', 'get', 'flowcollector', '-o', f'jsonpath="{{.items[*].spec.{agent}.cacheMaxFlows}}"']
+	}
+
+	# collect data from cluster about agent
+	for command in agent_commands:
+		info[command] = subprocess.run(agent_commands[command], capture_output=True, text=True).stdout[1:-1]
```
memodi, line 89: same comment as above for error checking.
> I would suggest to check stderr as well in case of errors before checking for stdout.

scripts/nope.py
```diff
@@ -59,8 +59,46 @@ def run_query(query):
 	return data.json()
 
 
+def get_operator_info():
```
memodi, line 62: suggest naming this function as: `get_netobserv_env_info` to resemble what it's doing specifically.

scripts/nope.py
```diff
@@ -59,8 +59,46 @@ def run_query(query):
...
+			print(f"\nExecuted agent command '{' '.join(agent_commands[command])}' to get {command} data")
+
+	# return all collected data
+	return info
+
+
 def main():
 
+	# get operator data
+	RESULTS["netobservEnv"] = get_operator_info()
```
memodi, line 100: since we're trying to have this tool extensible for other use cases as well. We should probably have an extra argument to fetch environment info and that argument could hold a `str` value, for e.g.: "netobserv". Internally in code, we could maintain a dictionary of all the environment that it supports with key as str and value as function name, for e.g.: {"netobserv":  "get_netobserv_env_info"}. So, whenever user wishes to grab particular env. info, appropriate function can be called based on user argument that was provided.

memodi (changes requested):
requesting changes for error handling and extensible use case.

scripts/nope.py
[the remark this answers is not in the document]
nathan-weinberg: The code will fail well before this step if KUBECONFG is not set - if we are going to do checks, it should be done prior to the argument parsing. I can introduce a function to do this validation, not necessarily related to this task but we can piggy-back it onto this PR.

scripts/nope.py
paigerube14, line 89: It looks like you are doing the same functionality a couple of times (calling subprocess and getting certain output), maybe this would be good in a function and call it instead

scripts/nope.py
nathan-weinberg: Let's hold off until doing this until we have an actual other usecase to test, I don't want to overengineer things in the meantime

paigerube14:
Looks good to me, not commenting to let Mehul verify the script itself

PR #177, opened 2022-06-07
make Netobserv commands NS scoped

memodi:
Make NetObserv commands NS scoped.

### DIFF
--- a/scripts/nope.py
+++ b/scripts/nope.py
@@ -106,9 +106,9 @@ def get_netobserv_env_info():
 	# intialize info and base_commands objects
 	info = {}
 	base_commands = {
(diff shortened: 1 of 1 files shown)


nathan-weinberg:
What's the motivation behind this change? Do we want to be hardcoding `network-observability` namespace? Is there any scenario in which this namespace might not be applicable?

memodi:
> What's the motivation behind this change? Do we want to be hardcoding `network-observability` namespace? Is there any scenario in which this namespace might not be applicable?

I just ran into error where this commands failed, where I ran the tool and my current context wasn't `network-observability` NS. There's no restriction on user to deploy other than `network-observability` , perhaps we can add CLI arg to specify NS but I just thought for now we should be okay to have it hard coded

PR #182, opened 2022-06-14
updates for ebpf, metrics analyze script and fixes

memodi:
1. Adding metrics and dashboard for ebpf agent.
2. Adding analyze script as an interim way to get some statistics easily in absence of ES uploads. Script runs on top the JSON file that nope.py tool outputs it can be simply run as: `./analyze_metrics.py <path to json file>`
3. fix in nope.py to just get `new-token` anyway instead of `get-token` so we don't run into errors.

### DIFF
--- a/scripts/analyze_metrics.py
+++ b/scripts/analyze_metrics.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import sys
 import json
--- a/scripts/analyze_metrics.py
+++ b/scripts/analyze_metrics.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3.9
+#!/usr/bin/env python
 
 import sys
 import json
--- a/scripts/analyze_metrics.py
+++ b/scripts/analyze_metrics.py
@@ -4,7 +4,7 @@ import sys
 import json
 import numpy as np
 
-IGNORE_DATA = ["benchmarkEnv", "netobservEnv"]
+IGNORE_DATA = ("benchmarkEnv", "netobservEnv")
 
 
 def compute_avg(resource, data):
--- a/scripts/analyze_metrics.py
+++ b/scripts/analyze_metrics.py
@@ -5,6 +5,24 @@ import json
 import numpy as np
 
 IGNORE_DATA = ("benchmarkEnv", "netobservEnv")
+AVERAGE_METRICS = (
+    "cpuUsage",
+    "memoryUsage",
+    "nFlowsProcessedPerMinute",
+    "FlowProcessingTime",
+    "nPacketsProcessedPerMinute",
+    "nFlowLogsProcessed",
+    "nBytes",
+    "ebpfCpuUsage",
+    "ebpfmemoryUsage",
+)
+PERC_90_METRICS = (
+    "nFlowsProcessedPerMinute",
+    "nPacketsProcessedPerMinute",
+    "FlowProcessingTime",
+    "nFlowLogsProcessed",
+    "nBytes",
+)
 
 
 def compute_avg(resource, data):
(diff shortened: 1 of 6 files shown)


scripts/analyze_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+        print(f"{metrics['metric']['pod']}, {str(perc_90th)}")
+
+
+def process_metrics(file):
+    with open(file) as fh:
+        metrics_data = json.load(fh)
+        for metric in metrics_data.keys():
+            if metric in IGNORE_DATA:
+                continue
+            if metric in (
```
nathan-weinberg, line 45: Can we use a list instead of a tuple here?

scripts/analyze_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+                "FlowProcessingTime",
+                "nPacketsProcessedPerMinute",
+                "nFlowLogsProcessed",
+                "nBytes",
+                "ebpfCpuUsage",
+                "ebpfmemoryUsage",
+            ):
+                compute_avg(metric, metrics_data[metric]["data"]["result"])
+
+            if metric in (
```
nathan-weinberg, line 58: Same comment as above

scripts/analyze_metrics.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env python3.9
```
nathan-weinberg, line 1:
```suggestion
#!/usr/bin/env python3
```

scripts/analyze_metrics.py
```diff
@@ -0,0 +1,73 @@
...
+                "nBytes",
+            ):
+                compute_90th_perc(metric, metrics_data[metric]["data"]["result"])
+
+            if metric == "lokiStorageUsage":
+                analyze_diskutil(metric, metrics_data[metric]["data"]["result"])
+
+
+if __name__ == "__main__":
+    file = sys.argv[1]
```
nathan-weinberg, line 72: Might want a basic check here just to ensure an arg is indeed passed

memodi: the `open` should fail if there are problems with file. I just added this script for us to use in interim until we do ES uploads at which point this script will be of no use,  it's definitely not at a level where it could be distributed and I might just remove it later from the repo.

scripts/analyze_metrics.py
memodi: python 3.6 and later is needed here since I am using f strings in formatting.

nathan-weinberg: Right, but this ties us explicitly to 3.9. We already have a version specified in the README so I'd rather not complicate the pathing/versioning.

scripts/analyze_metrics.py
memodi: again just a condition check

scripts/analyze_metrics.py
memodi: nvm, I modified it to be global tuples for better readability.

scripts/analyze_metrics.py
memodi: nvm, I modified it to be global tuples for better readability.

scripts/analyze_metrics.py
memodi: Okay, done. updated in recent commits.

scripts/analyze_metrics.py
nathan-weinberg: the first `(` here is the beginning of a tuple, I'm saying use a list instead

nathan-weinberg: You've since moved the tuple to a constant above

scripts/analyze_metrics.py
nathan-weinberg: See above

scripts/analyze_metrics.py
memodi: any particular reason why list is preferred here over tuple? Reason I thought tuple would be better choice since we're not looking to modify that  data.

nathan-weinberg: I typically use lists over tuples in general, but I suppose that is true. Won't block merge on this

PR #186, opened 2022-06-16
Adding architecture type to cluster scaling 

paigerube14:
Need to see the architecture type to properly add the infrastructure and workload nodes of the right types 

https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/scaling-arm/13/console
The infra and workload nodes got added correctly. The dittybopper pod just didn't go running in time. Will see if we should increase the wait time on that

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,7 +23,7 @@ pipeline {
         3.4~3.7: ansible-2.4-extra || ansible-2.3 <br/>
         '''
         )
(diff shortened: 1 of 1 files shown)


qiliRedHat:
@paigerube14 Thanks for the quick fix. I've approved it so we are unblocked. For dittybopper, I guess dittybopper images may not be compatible with arm.

PR #193, opened 2022-06-29
Adding osp call to cluster-post-config

paigerube14:
Need to be able to call the cluster-post-config in cluster-workers-scaling and properly pass the OSP parameters needed

Passing job: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/osp_scale/2/console

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -133,6 +133,10 @@ pipeline {
                 node_name=$(oc get node --no-headers | grep master| head -1| awk '{print $1}')
                 oc get node $node_name -ojsonpath='{.status.nodeInfo.architecture}'
               ''')
+              ENV_VARS += '''
+OPENSHIFT_PROMETHEUS_RETENTION_PERIOD=15d
+OPENSHIFT_PROMETHEUS_STORAGE_SIZE=500Gi
+OPENSHIFT_ALERTMANAGER_STORAGE_SIZE=20Gi'''
               println "architecture_type $architecture_type"
               if(env.VARIABLES_LOCATION.indexOf("aws") != -1){
                 if (architecture_type.contains("arm64")) {
@@ -158,9 +162,6 @@ OPENSHIFT_INFRA_NODE_VOLUME_TYPE=gp2
 OPENSHIFT_INFRA_NODE_VOLUME_SIZE=100
 OPENSHIFT_PROMETHEUS_STORAGE_CLASS=gp2
 OPENSHIFT_ALERTMANAGER_STORAGE_CLASS=gp2
-OPENSHIFT_PROMETHEUS_RETENTION_PERIOD=15d
-OPENSHIFT_PROMETHEUS_STORAGE_SIZE=500Gi
-OPENSHIFT_ALERTMANAGER_STORAGE_SIZE=20Gi
 OPENSHIFT_WORKLOAD_NODE_VOLUME_IOPS=0
 OPENSHIFT_WORKLOAD_NODE_VOLUME_TYPE=gp2
 OPENSHIFT_WORKLOAD_NODE_VOLUME_SIZE=500
@@ -180,9 +181,6 @@ OPENSHIFT_INFRA_NODE_VOLUME_TYPE=Premium_LRS
 OPENSHIFT_INFRA_NODE_VM_SIZE=Standard_D48s_v3
 OPENSHIFT_PROMETHEUS_STORAGE_CLASS=managed-csi
 OPENSHIFT_ALERTMANAGER_STORAGE_CLASS=managed-csi
-OPENSHIFT_PROMETHEUS_RETENTION_PERIOD=15d
-OPENSHIFT_PROMETHEUS_STORAGE_SIZE=500Gi
-OPENSHIFT_ALERTMANAGER_STORAGE_SIZE=20Gi
 OPENSHIFT_WORKLOAD_NODE_VOLUME_SIZE=500
 OPENSHIFT_WORKLOAD_NODE_VOLUME_TYPE=Premium_LRS
 OPENSHIFT_WORKLOAD_NODE_VM_SIZE=Standard_D32s_v3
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -247,7 +247,17 @@ OPENSHIFT_PROMETHEUS_RETENTION_PERIOD=15d
 OPENSHIFT_PROMETHEUS_STORAGE_SIZE=500Gi
 OPENSHIFT_ALERTMANAGER_STORAGE_SIZE=20Gi
               ''')]
-            } else {
+            } else if(env.VARIABLES_LOCATION.indexOf("osp") != -1){
+              build job: 'scale-ci/e2e-benchmarking-multibranch-pipeline/cluster-post-config', parameters: [
+              string(name: 'BUILD_NUMBER', value: BUILD_NUMBER), string(name: 'HOST_NETWORK_CONFIGS', value:'false'),
```
nathan-weinberg, line 252: Two vars on the same line here
```suggestion
              string(name: 'BUILD_NUMBER', value: BUILD_NUMBER),
              string(name: 'HOST_NETWORK_CONFIGS', value:'false'),
```

nathan-weinberg:
LGTM, but I'll let others comment before merging

qiliRedHat:
1. About the vm type, for AWS, we used
INFRA m5.12xlarge 48vCPU 192GiB
WORKLOAD m5.4xlarge 16vCPU 64GiB

For OSP, the ci.cpu types' memory are small while cpu are big.
INFRA ci.cpu.xxxl 48vCPU 20 GB
WORKLOAD ci.cpu.xxl 36vCPU 20 GB

For OSP, considering we run on 20 nodes on OSP, not as large as AWS (120 nodes), I think the following types may be better.  Surely we have to consider the quota for the types too.
INFRA
ci.m5.xlarge	24	48 GB
WORKLOAD
ci.m5.large	16vCPU	32 GB 


2. I didn't see the following envs passed to OSP.
```
OPENSHIFT_PROMETHEUS_RETENTION_PERIOD=15d
OPENSHIFT_PROMETHEUS_STORAGE_SIZE=500Gi
OPENSHIFT_ALERTMANAGER_STORAGE_SIZE=20Gi
```
While this line of code is run in the [test job](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/cluster-post-config/532/console) 
https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/0cdb1cf87c95fc990e030742426df90cc6903aa9/Jenkinsfile#L407

```
06-29 22:05:50.595  ++ grep vsphere -c
06-29 22:05:50.595  + [[ 0 > 0 ]]
06-29 22:05:50.595  + envsubst
06-29 22:05:50.595  + oc apply -f -
06-29 22:05:54.748  configmap/cluster-monitoring-config created
06-29 22:05:54.748  + oc patch -n openshift-ingress-operator ingresscontrollers.operator.openshift.io default -p '{"spec": {"nodePlacement": {"nodeSelector": {"matchLabels": {"node-role.kubernetes.io/infra": ""}}}}}' --type merge
06-29 22:05:54.748  ingresscontroller.operator.openshift.io/default patched
```
I wonder what's the status of the monitoring pods after the run.
If OSP can support the persistent storage for Monitoring, we should pass the ENVs. If not, OSP should be like the vshpere, moving the monitoring to infra nodes but doesn't configure the persistent storage.

paigerube14:
I think we are going to have to experiment with the sizing of infra and workload nodes to see if we need the CPU or storage more. We could do what you suggested with not moving the storage to the infra/workload nodes but I just wonder if we will notice a difference. Think we need to experiment with large clusters and see how they perform if we leave it as is and then with not moving the storage. Thoughts?

I do not think any of the clouds currently are passing the PROMETHEUS type env vars. I will add those for this PR 

Looking at the configuration sheet it looks like the extra-small and small profiles (3/20) are the same sizing as AWS but there's no information for medium sizing for OSP. Why do you say we are only running up to 20 nodes?

paigerube14:
Properly passing the prometheus variables at the start of the specific cloud types working here: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/osp_scale/4/console

qiliRedHat:
> I think we are going to have to experiment with the sizing of infra and workload nodes to see if we need the CPU or storage more. We could do what you suggested with not moving the storage to the infra/workload nodes but I just wonder if we will notice a difference. Think we need to experiment with large clusters and see how they perform if we leave it as is and then with not moving the storage. Thoughts?

I'm worried about the 20G memory, as in bug https://bugzilla.redhat.com/show_bug.cgi?id=2103048, with persistent storage configured, the promethues-k8s pods memory usage grew up to >20G in router-perf test. I guess if the infra nodes have only 20g memory, it could become NotReady.

What do you mean by `with not moving the storage to the infra/workload nodes`?

> Looking at the configuration sheet it looks like the extra-small and small profiles (3/20) are the same sizing as AWS but there's no information for medium sizing for OSP. Why do you say we are only running up to 20 nodes?

Because I saw Roshni planed 20 nodes in our manual scale-ci test https://issues.redhat.com/browse/OCPQE-10768.

paigerube14:
@qiliRedHat I updated the infra and workload node types to what you suggested

You had suggested: `OSP should be like the vshpere, moving the monitoring to infra nodes but doesn't configure the persistent storage.` do we still want to do that or try with the certain node sizes you had suggested and see how it works? 

The jira story you linked is a specific type of osp testing I am supposed to do. We shouldn't have any issues scaling up to 120 nodes just like other cloud types. I thought we had done that configuration before but I can not find the results anywhere.

qiliRedHat:
@paigerube14
`If OSP can support the persistent storage for Monitoring, we should pass the ENVs. If not, OSP should be like the vshpere, moving the monitoring to infra nodes but doesn't configure the persistent storage.`
As you added the persistent storage related ENVs to all clouds, that should not be a problem.

PR #204, opened 2022-07-11
Adding router perf cerberus check and csv_gen

paigerube14:
Adding in a boolean cerberus check at the end of the router_perf test
Also adding gen_csv to generate the google sheet from benchmark-comparison

https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/router-perf/14/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -141,10 +141,13 @@ pipeline {
                 def result = cerberus_job.result.toString()
                 println "cerberus result $result"
                 if( status == "PASS") {
+                    println "previous status = pass"
                     if (cerberus_job.result.toString() != "SUCCESS") {
                         status = "Cerberus check failed"
+                        currentBuild.result = "FAILURE"
                     }
                 } else {
+                    println "previous test had already failed"
                     if (cerberus_job.result.toString() != "SUCCESS") {
                         status += ", Cerberus check failed"
                         currentBuild.result = "FAILURE"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,13 +8,14 @@ if (userId) {
 
 def RETURNSTATUS = "default"
 def output = ""
-
+def status = ""
 pipeline {
   agent none
 
   parameters {
         string(name: 'BUILD_NUMBER', defaultValue: '', description: 'Build number of job that has installed the cluster.')
         booleanParam(name: 'WRITE_TO_FILE', defaultValue: false, description: 'Value to write to google sheet (will run https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/write-scale-ci-results)')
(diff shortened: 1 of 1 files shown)


qiliRedHat:
@paigerube14 Thanks for adding these features to router-perf!
I have some concerns. 
1. The failed [cerberus job ](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/cerberus/94/console) didn't make the router-perf test failed. Is this expected and the same behavior with kube-burner now?
2. The cerberus job failed for `Failed containers in installer-7-ip-10-0-172-215.us-east-2.compute.internal: ['installer']`, I think this is a false negative in most of the case. Can cerberus be configured to exclude these kind of pod's check?
3. Is the new added `GEN_CSV` parameter used in the later code? router-perf already has it's g-sheet generation code here https://github.com/cloud-bulldozer/e2e-benchmarking/blob/faf70c828056806eb9f26c62094ba2e80313a3ab/workloads/router-perf-v2/common.sh#L260.

paigerube14:
@qiliRedHat 

1. Yes it should've failed because of the cerberus failure, I will look into that today and update this PR 

2. We need to figure out a list of regex style namespaces to look at and pass to cerberus. I wanted our list to include namespaces that were created from any scale-ci run but also almost all openshift specific namespaces in case the test effects any areas we aren't checking. The installer one we should ignore and I think the community operators refresh every couple of minutes that the pod fails and could again be a false negative failure

3. With this newest [PR](https://github.com/cloud-bulldozer/e2e-benchmarking/pull/441) the generation of a sheet is not default anymore, see also this [env var ](https://github.com/cloud-bulldozer/e2e-benchmarking/blob/faf70c828056806eb9f26c62094ba2e80313a3ab/workloads/router-perf-v2/env.sh#L49)
You can see here it creates a table layout now in the [router-perf job ](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/router-perf/610/console) instead of a google sheet

qiliRedHat:
Thanks for the information Paige. 
2. `Currently there is no way in cerberus to ignore certain pods in a namespace`. Could it possible to add some code to the cerberus Jenkins job to generate a successful result if only the 'installer' issue is reported by cerberus and it is a false negative.? It's not a hurry. Another option is to list the problem namespace's pods with name 'installer' to help us to confirm if it is a real false negative - helpful to the cluster that are destroyed right after test.
3. Thanks for sharing the new update. If not adding the GEN_CSV and set to true, google sheet is not generated. I hope g-sheet is to be generated by default as before. Thanks for adding that!

qiliRedHat:
> @qiliRedHat definitely can add a check for if the pod has "installer" in the name. Just don't want to form a long list of other pods to ignore just to get a passing result
Thanks, we can do it in another pr. Let's merge this one first.

qiliRedHat:
Last /approve didn't get this merged

PR #205, opened 2022-07-11
Network perf cerberus check

paigerube14:
Adding in cerberus check into network-perf
Also setting to be able to generate google sheet

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,6 +8,8 @@ if (userId) {
 
 def RETURNSTATUS = "default"
 def output = ""
+def status = "FAIL"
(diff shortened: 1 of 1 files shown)


paigerube14:
There was an issue with the changing of directories that it couldn't find the run script. Should be good now: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/network-perf/44/console

qiliRedHat:
@paigerube14 Sorry for the delay. 
/lgtm

PR #210, opened 2022-07-19
Adding missing publicLoadBalancer to Azure machinesets

skordas:
Due issues with pulling images from quay.io ${TITLE}


qiliRedHat:
@skordas Is publicLoadBalancer also in a NON private cluster's worker machinesets?

skordas:
@paigerube14 PTAL

PR #212, opened 2022-07-27
Fixes and enhancements to NOPE tool

nathan-weinberg:
Summary of changes:
- Rearranged parsing order of NOPE fields to make output more readable
- Added ES_USERNAME and ES_PASSWORD as environmental variables to be set
- Altered data schema to improve ES compatibility
- Dumped JSON files now show if data is full or partial
- Fixed bug with boolean fields in Jenkins params breaking ES upload
- Fixed bug with Jenkins params only being collected sometimes
- Renamed some files for increased clarity
- Updated README


scripts/nope.py
```diff
@@ -28,55 +29,62 @@
...
+	# intialize post-processing data objects
+	metadata = {"query": query}
+	clean_data = []
 
 	try:
-		# add metadata artifacts from prometheus return object
-		for k, v in raw_data["data"]["result"][0]["metric"].items():
-			clean_data["metadata"][k] = v
+		# capture metadata artifacts from prometheus return object
+		for metadata_point_name, metadata_point_value in raw_data["data"]["result"][0]["metric"].items():
```
memodi, line 66: this is the source of bug why only one single pod cpu/memory usage is uploaded. The problem is: “result” returned from Thanos query is list of dictionaries. I tested out and below code works as expected for dumping data to JSON correctly. Have not tested ES upload with this code though:
```python3        
      for result in raw_data["data"]["result"]:
            for data_point in result["values"]:
                clean_data.append(
                    {
                        "uuid": UUID,
                        "metric_name": metric_name,
                        "data_type": "datapoint",
                        "unix_timestamp": data_point[0],
                        "value": float(data_point[1]),
                        "metdata": result["metric"],
                    }
                )

File: scripts/nope.py [hunk 2]
```diff
@@ -28,55 +29,62 @@
...
+			metadata[metadata_point_name] = metadata_point_value
 
-		# add value artifacts from prometheus return object
+		# capture value artifacts from prometheus return object and add previously collected metadata
 		for data_point in raw_data["data"]["result"][0]["values"]:
```
memodi, line 70: another source of bug why only one single pod cpu/memory usage is uploaded. this for loop should not be needed if proposed code in above comment is added.

memodi (changes requested):
Besides the added comments, highly recommend to use spaces instead of tabs in python3 code as per pep8 styling. It resulted in error when attempted to modify code.

scripts/nope.py
nathan-weinberg: Implemented and working with ES! Thanks for the fix!

scripts/nope.py
nathan-weinberg: Removed as part of the above fix!

memodi (approved):

PR #215, opened 2022-08-02
Reconfigure Netobserv perf job to do install and config of NOO

nathan-weinberg:
Resolves [NETOBSERV-499](https://issues.redhat.com/browse/NETOBSERV-499)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,200 +1,214 @@
 @Library('flexy') _
 
 // rename build
-def userId = currentBuild.rawBuild.getCause(hudson.model.Cause$UserIdCause)?.userId ?: "memodi"
+def userId = currentBuild.rawBuild.getCause(hudson.model.Cause$UserIdCause)?.userId
 if (userId) {
   currentBuild.displayName = userId
 }
 
 pipeline {
-  agent none
 
-  parameters {
-        string(name: 'BUILD_NUMBER', defaultValue: '', description: 'Build number of job that has installed the cluster.')
-        string(name:'JENKINS_AGENT', defaultValue:'goc48',description:
-        '''
-        scale-ci-static: for static agent that is specific to scale-ci, useful when the jenkins dynamic agen
- isn't stable<br>
-        4.y: oc4y || mac-installer || rhel8-installer-4y || <br/>
-            e.g, for 4.8, use oc48 || mac-installer || rhel8-installer-48 <br/>
-            for agents with go tools, use goc48, goc49 etc. <br/> 
-        '''
+    agent { label params.JENKINS_AGENT_LABEL }
+
+    options {
+        timeout(time: 1, unit: 'HOURS')
+    }
+
+    parameters {
+        string(
+            name: 'JENKINS_AGENT_LABEL',
+            defaultValue:'oc411',
+            description: 'Label of Jenkins agent to execute job'
         )
-
-        choice(name: 'Deploy_NetObserv', choices: ['OperatorHub', 'main', 'None'], description: '''Option to Deploy netobserv from main or OperatorHub release, select one.<br>
-            If none selected tests will run without NetObserv deployed <br/>''')
-
-        text(name: 'ENV_VARS', defaultValue: '', description:'''<p>
-               Enter list of additional (optional) Env Vars you'd want to pass to the script, one pair on each line. <br>
-               e.g.<br>
-               SOMEVAR1='env-test'<br>
-               SOMEVAR2='env2-test'<br>
-               ...<br>
-               SOMEVARn='envn-test'<br>
-               </p>'''
-            )
-        string(name: 'E2E_BENCHMARKING_REPO', defaultValue:'https://github.com/memodi/e2e-benchmarking', description:'You can change this to point to your fork if needed. Upstream is: https://github.com/cloud-bulldozer/e2e-benchmarking')
-        string(name: 'E2E_BENCHMARKING_REPO_BRANCH', defaultValue:'netobserv-trials', description:'You can change this to point to a branch on your fork if needed. Change it to "master" if using upstream repo.')
-        choice(name: 'WORKLOAD_TYPE', choices: ['uperf', 'node-density-heavy'], description: 'Specify workload type')
-
-        // netobserv params
-        choice(name: 'FLOW_SAMPLING', choices: ['1', '100', '400'], description: 'Specify flow sampling rate')
-
-        // uperf-only params
-        separator(name: "UPERF_WORKLOAD", sectionHeader: "UPERF Job Options", separatorStyle: "border-width: 0",
-        sectionHeaderStyle: """
-				font-size: 20px;
-				font-weight: bold;
-				font-family: 'Orienta', sans-serif;""")
-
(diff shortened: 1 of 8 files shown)


memodi (approved):
@nathan-weinberg - just one minor comment on something if we want to include in this PR while we're at it.

scripts/netobserv.sh
```diff
@@ -74,4 +74,5 @@ uninstall_operatorhub_netobserv() {
   oc delete -f $WORKSPACE/ocp-qe-perfscale-ci/scripts/noo-subscription.yaml
   oc delete csv/netobserv-operator.v0.1.3 -n network-observability
```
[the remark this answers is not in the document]
nathan-weinberg: Sure, what is the change that's needed?

scripts/README.md
```diff
@@ -52,12 +52,12 @@ export THANOS_URL=https://`oc get route thanos-querier -n openshift-monitoring -
 
 ### Updating common parameters of flowcollector
 You can update common parameters of flowcollector with the following commands:
-- **IPFix sampling rate:** `$ oc patch flowcollector cluster --type=json -p '[{"op": "replace", "path": "/spec/ipfix/sampling", "value": <value>}]'`
-- **CPU limit:** `$ oc patch flowcollector  cluster --type=json -p '[{"op": "replace", "path": "/spec/flowlogsPipeline/resources/limits/cpu", "value": "<value>m"}]'`
+- **IPFix sampling rate:** `$ oc patch flowcollector cluster --type=json -p "[{"op": "replace", "path": "/spec/ipfix/sampling", "value": <value>}]"`
```
paigerube14, line 55: do these lines work with those quotations? Since you have double quote on the outside and inside I'm not sure if they'll work

scripts/run_uperf.sh
```diff
@@ -4,17 +4,17 @@ CR=ripsaw-uperf-crd.yaml
 # run network-perf common.sh
 cd $NETWORK_PERF_DIR && source common.sh && cd -
 
-log "###############################################"
-log "Workload: ${WORKLOAD}"
-log "Network policy: ${NETWORK_POLICY}"
-log "Samples: ${SAMPLES}"
-log "Pairs: ${PAIRS}"
+echo "###############################################"
```
paigerube14, line 7: @nathan-weinberg is there a specific reason you're not just calling these workload job jenkins jobs specifically. Think having specific scripts here will cause us to get out of date very quickly

Jenkinsfile
```diff
@@ -1,200 +1,214 @@
...
-        sectionHeaderStyle: """
-				font-size: 20px;
-				font-weight: bold;
-				font-family: 'Orienta', sans-serif;""")
-        string(name: 'NODE_COUNT', defaultValue: '3', description: 'Number of nodes to be used in your cluster for this workload. Should be the number of worker nodes on your cluster')
-        string(name: 'VARIABLE', defaultValue: '100', description: 'This will export PODS_PER_NODE env variable; set to 200, work up to 250. Creates this number of applications proportional to the calculated number of pods / 2 Read here for detail of each variable')
+    agent { label params.JENKINS_AGENT_LABEL }
 
-        // TBA: http router params
+    options {
```
paigerube14, line 13: Will this job always finish within an hour? that seems a little low to me if we have large clusters

paigerube14: Is this only doing the calls to elastic search not actually running a workload? Might be misunderstanding the code here

Jenkinsfile
```diff
@@ -1,200 +1,214 @@
...
+                copyArtifacts(
+                    fingerprintArtifacts: true, 
+                    projectName: 'ocp-common/Flexy-install',
+                    selector: specific(params.FLEXY_BUILD_NUMBER),
+                    target: 'flexy-artifacts'
+                )
+                checkout([
+                    $class: 'GitSCM',
+                    branches: [[name: 'netobserv-perf-tests' ]],
+                    userRemoteConfigs: [[url: 'https://github.com/openshift-qe/ocp-qe-perfscale-ci' ]],
```
paigerube14, line 99: would be good to have this set from the current branch and git url that is being used when this jenkins file is run. See [here](https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/73b85d2d567acd465fd40f29f17e7beae8e179ae/Jenkinsfile#L55) as an example. This way we can always know that the code in our specific branch or fork is being properly run

scripts/README.md
nathan-weinberg: Yes they do

scripts/run_uperf.sh
nathan-weinberg: @paigerube14 these are old scripts from @memodi, I just did a batch update of everything from `log` to `echo` - this won't be used in the job

Jenkinsfile
nathan-weinberg: Right now yes - I'll extend the timeout when the additional functionality is in place

scripts/run_uperf.sh
paigerube14: should we go ahead and delete these files then so we don't get confused

Jenkinsfile
nathan-weinberg: So how do these keywords work @paigerube14? Are they built-in? What do they point to?

If I can just update to these keywords, agreed!

scripts/run_uperf.sh
nathan-weinberg: Up to @memodi - we might want to keep them for manual work if Jenkins is down, but personally I don't feel strongly one way or the other.

scripts/netobserv.sh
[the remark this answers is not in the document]
memodi:
```suggestion
oc delete csv -l operators.coreos.com/netobserv-operator.network-observability```

File: Jenkinsfile (hunk 5 above)
Comment (paigerube14) [reply Jenkinsfile:99]: you can just update to those keywords. It will point to the git url that the current repo it's on with the current branch. So in your case if you ran this jenkins branch under your forked e2e-benchmark jenkins job it would have GIT_BRANCH=jenkins and GIT_URL=https://github.com/nathan-weinberg/ocp-qe-perfscale-ci.git 

this way code will get checked out based on your current branch so we can get the most up to date code and can validate it works before pushing into the main branches

File: scripts/run_uperf.sh (hunk 3 above)
Comment (memodi) [reply scripts/run_uperf.sh:7]: @paigerube14 @nathan-weinberg  it's alright to delete them since we're moving towards different way to run workloads. Don't wanna block this PR though for that, I'll create a PR later to clean this up.

File: OWNERS [hunk 6]
```diff
@@ -1,14 +1,13 @@
...
 reviewers:
-- jechen0648
-- qiliRedHat
 - mffiedler
-- paigerube14
-- rpattath
-- skordas
 - memodi
 - nathan-weinberg
+- paigerube14
```
paigerube14, line 9: do you want to add me to an approver?

paigerube14:
looks god to me but going to let Mehul do final approval

OWNERS
nathan-weinberg: Yes! Thanks!

PR #224, opened 2022-08-16
Removed ipfix references from automation

nathan-weinberg:
Resolves [NETOBSERV-528](https://issues.redhat.com/browse/NETOBSERV-528)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -232,6 +232,7 @@ pipeline {
                     // attempt updating common parameters of flowcollector
                     println 'Updating common parameters of flowcollector...'
                     returnCode = sh(returnStatus: true, script: """
+                        oc patch flowcollector cluster --type=json -p "[{"op": "replace", "path": "/spec/agent", "value": "ebpf"}] -n network-observability"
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -237,8 +232,7 @@ pipeline {
                     // attempt updating common parameters of flowcollector
                     println 'Updating common parameters of flowcollector...'
                     returnCode = sh(returnStatus: true, script: """
-                        oc patch flowcollector cluster --type=json -p "[{"op": "replace", "path": "/spec/agent", "value": ${params.COLLECTOR_AGENT}}] -n network-observability"
```
memodi, line 240: we would need this patch until upstream [flowcollector CRD](https://raw.githubusercontent.com/netobserv/network-observability-operator/main/config/samples/flows_v1alpha1_flowcollector_versioned.yaml) is updated which is used [here](https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/523cac2c9bd62f9c44bde3cfa60c7342a6b2faaf/scripts/netobserv.sh#L12 )

memodi (approved):

memodi (changes requested):
requesting changes to restore patch, could use hardcoded ebpf agent in there.

Jenkinsfile
nathan-weinberg: Done

PR #226, opened 2022-08-16
Adding check if file types exist and cerberus cleanup params

paigerube14:
Fixing: 
* Adding params for cerberus and cleanup to pass to sub jobs
* Adding in file check for earlier release that don't have _flexy-artifacts/workdir/install-dir/cluster_info.yaml_ file 
* Spacing out parameters

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,65 +26,138 @@ pipeline{
     agent any
 
     parameters {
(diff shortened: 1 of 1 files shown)


qiliRedHat:
@paigerube14 /lgtm

> Adding params for cerberus and cleanup to pass to sub jobs

For cron triggered job, we can do the same

PR #227, opened 2022-08-17
Post config arm arch example 

paigerube14:
Adding example configurations for arm aws types

cc: @svetsa-rh


paigerube14:
/assign @skordas 
Fixed based on comments from slack

mffiedler:
Go ahead and squash and I will approve

PR #230, opened 2022-08-18
adding upstream user when running from loaded-upgrade jobs and cerber…

qiliRedHat:
adding upstream user when running from loaded-upgrade jobs
add cerberus pod ignore pattern

- https://github.com/redhat-chaos/cerberus/pull/178

Update router-perf default config to avoid connection error.

- https://github.com/cloud-bulldozer/e2e-benchmarking/issues/451
- https://github.com/cloud-bulldozer/e2e-benchmarking/pull/452

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,11 +1,27 @@
 @Library('flexy') _
 
 // rename build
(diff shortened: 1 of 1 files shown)


qiliRedHat:
@paigerube14 PTAL

PR #232, opened 2022-08-19
fix typo

qiliRedHat:
@SachinNinganure PTAL


qiliRedHat:
@paigerube14 PTAL
I think we can add more team members to reviewer list.

PR #240, opened 2022-08-26
kafka support

memodi:
Adding Kafka support in scripts and Jenkinsfile. [NETOBSERV-523](https://issues.redhat.com/browse/NETOBSERV-523) and [NETOBSERV-524](https://issues.redhat.com/browse/NETOBSERV-524)
[Success run](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/memodi-e2e-benchmarking-multibranch-pipeline/job/kafka-update/8/console)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -90,6 +90,11 @@ pipeline {
             '''
         )
         booleanParam(
+            name: 'ENABLE_KAFKA',
+            defaultValue: false,
+            description: 'Check this box to setup Kafka for NetObserv'
+        )
+        booleanParam(
             name: 'USER_WORKLOADS',
             defaultValue: true,
             description: 'Check this box to setup FLP service and create service-monitor'
@@ -274,6 +279,24 @@ pipeline {
                     }
                     else {
                         println 'Successfully updated common parameters of flowcollector :)'
+                    }
+                    println "Checking if Kafka needs to be enabled.."
+                    if (env.ENABLE_KAFKA == "true") {
+                        returnCode = sh(returnStatus: true, script:"""
+                            echo "Enabling Kafka in flowcollector"
+                            oc patch flowcollector cluster --type=json -p "[{"op": "replace", "path": "/spec/kafka/enable", "value": "true"}] -n network-observability"
+                            source $WORKSPACE/ocp-qe-perfscale-ci/scripts/netobserv.sh
+                            deploy_kafka
+                        """)
+                        if (returnCode.toInteger() != 0) {
+                            error('Failed to enable Kafka in flowcollector')
+                        }
+                        else {
+                            println 'Successfully enabled Kafka with flowcollector'
+                        }
+                    }
+                    else {
+                        println "Skipping Kafka deploy"
                     }
                 }
             }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -280,23 +280,23 @@ pipeline {
                     else {
                         println 'Successfully updated common parameters of flowcollector :)'
                     }
-                    println "Checking if Kafka needs to be enabled.."
-                    if (env.ENABLE_KAFKA == "true") {
+                    println "Checking if Kafka needs to be enabled..."
+                    if (params.ENABLE_KAFKA == true) {
                         returnCode = sh(returnStatus: true, script:"""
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -275,6 +280,24 @@ pipeline {
...
+                        if (returnCode.toInteger() != 0) {
+                            error('Failed to enable Kafka in flowcollector')
+                        }
+                        else {
+                            println 'Successfully enabled Kafka with flowcollector'
+                        }
+                    }
+                    else {
+                        println "Skipping Kafka deploy"
+                    }
```
nathan-weinberg, line 300: I think this should be its own stage, something like the following
```groovy
stage(Enable Kafka) {
    when {
        expression {params.ENABLE_KAFKA == true}
    }
    steps {
        ...
    }
}
```

nathan-weinberg (changes requested):

Jenkinsfile
nathan-weinberg: Unless this is flowcollector-specific @memodi? I just may not be understanding the architecture

memodi: It can certainly be in its own stage. I debated a bit but since we're patching flowcollector again here and TBH was bit hesitant on adding a one more stage just for Kafka, I figured it to include in the same stage as "Update flowcollector params" since we're patching here again to enable Kafka. But I can separate this out in its own stage if you think that's better suited.

nathan-weinberg: Since we are patching flowcollector I think it's fine have it been part of the existing stage :+1:

Jenkinsfile
```diff
@@ -89,6 +89,11 @@ pipeline {
                 1x.medium - Nodes size >= m5.8xlarge<br/>
             '''
         )
+        booleanParam(
+            name: 'ENABLE_KAFKA',
+            defaultValue: false,
+            description: 'Check this box to setup Kafka for NetObserv'
+        )
```
nathan-weinberg, line 96: Let's move this under `REPLICAS` so the template order matches the Groovy actions

Jenkinsfile
nathan-weinberg, line 300:
```suggestion
                    println "Checking if Kafka needs to be enabled..."
                    if (params.ENABLE_KAFKA == true) {
                        returnCode = sh(returnStatus: true, script:"""
                            echo "Enabling Kafka in flowcollector..."
                            oc patch flowcollector cluster --type=json -p "[{"op": "replace", "path": "/spec/kafka/enable", "value": "true"}] -n network-observability"
                            source $WORKSPACE/ocp-qe-perfscale-ci/scripts/netobserv.sh
                            deploy_kafka
                        """)
                        if (returnCode.toInteger() != 0) {
                            error('Failed to enable Kafka in flowcollector :(')
                        }
                        else {
                            println 'Successfully enabled Kafka with flowcollector :)'
                        }
                    }
                    else {
                        println "Skipping Kafka deploy..."
                    }
```

nathan-weinberg (changes requested):

Jenkinsfile
memodi: Sure, I'll move, but can you elaborate more about "Groovy actions"?

nathan-weinberg: Yes - so right now the order of the template params largely matches when they are used/executed upon in the Groovy code. So since Kafka is an optional flag that will run after other flowcollector config, I think it makes sense to have the checkbox follow those configs in the template. I am flexible however if you would prefer having it somewhere else!

memodi: Ah, that make sense. I moved it after REPLICAS however on rebuilding [multibranch pipeline](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/memodi-e2e-benchmarking-multibranch-pipeline/job/kafka-update/build?delay=0sec) on my branch, it would not re-order on UI, perhaps it will show correctly on e2e netobserv-perf-tests pipeline on merge?

nathan-weinberg: Yes, should be fine, I won't block on this :+1:

PR #258, opened 2022-09-12
Updated default Loki bucket name

nathan-weinberg:
Also updated a couple of `when` conditions

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -440,7 +440,7 @@ pipeline {
         }
         stage('Run NOPE tool') {
             when {
-                expression { params.NOPE == true }
+                expression { params.WORKLOAD != 'None' && params.NOPE == true }
             }
             steps {
                 copyArtifacts(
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -490,7 +490,7 @@ pipeline {
         }
         stage('Run Touchstone tool') {
             when {
-                expression { params.NOPE == true && params.NOPE_DUMP == false && currentBuild.currentResult != "UNSTABLE" }
+                expression { params.WORKLOAD != 'None' && params.NOPE == true && params.NOPE_DUMP == false && currentBuild.currentResult != "UNSTABLE" }
```
memodi, line 493: for testing changes faster in nope.py and touchstone, we would need to test them locally and cannot be tested on Jenkins without running workloads, which should be okay but just calling them out here.

memodi (approved):
/lgtm

Jenkinsfile
memodi: that's right, I forget about the piece where `./nope.py` start and end times are actually derived from workload runs.

PR #264, opened 2022-09-15
Default storage class 

paigerube14:
### Issue
Monitoring pods were not going ready/running because they made reference to a storage class that did not exist on the cluster. 
I added in a way to verify if the storage class existed on the cluster and if not set the prometheus/monitoring storage class to be the default storage class found in the cluster

There was also an issue that we never fully checked that the monitoring pods were running before marking the job successful. I added in code to make sure the statefulset that creates the pods has 2 replicas and the 2 running monitoring pods are on the infra nodes

Not setting new storage class and still successful 
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/default_storage_class/14/consoleFull

Successful job of waiting for setting new storage class than one given for prometheus, moving the monitoring pods to the infra nodes and go running https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/default_storage_class/19/console
**setting new storage classes to gp3-csi**

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -278,11 +278,11 @@ pipeline {
           currentBuild.description = "Copying Artifact from Flexy-install build <a href=\"${buildinfo.buildUrl}\">Flexy-install#${params.BUILD_NUMBER}</a>"
           buildinfo.params.each { env.setProperty(it.key, it.value) }
         }
-        ansiColor('xterm') {
+        script {
         withCredentials([file(credentialsId: 'b73d6ed3-99ff-4e06-b2d8-64eaaf69d1db', variable: 'OCP_AWS'),
                          file(credentialsId: 'eb22dcaa-555c-4ebe-bb39-5b25628cc6bb', variable: 'OCP_GCP'),
                          file(credentialsId: 'ocp-azure', variable: 'OCP_AZURE')]) {
-            sh label: '', script: '''
+            def RETURNSTATUS = sh(returnStatus: true, script: '''
               # Get ENV VARS Supplied by the user to this job and store in .env_override
               echo "$ENV_VARS" > .env_override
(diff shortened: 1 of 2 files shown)


qiliRedHat:
@paigerube14 Code looks good to me. I like you move it to a shell script, is very clear.
This recalled me another requirement. It can be done in an other PR later or if you like add to this one.
In our 4.11 retrospective meeting, Sharada said when there is a problem happened in this job,  if the previous job had finished partially, for example, machinesets are already installed, but moving monitoring had problem, rerun the job could fail because machinesets are already installed. I think we have the below options, let me know your thoughts.
1. do a cleanup of machinesets at the beginning
2. do a roll back at the end if some steps failed
3. provide check boxes to enable/disable steps separately (install infra/install workload/ move ingress/move monitoring)
CC: @svetsa-rh

paigerube14:
@qiliRedHat @svetsa-rh In this job, the infra/workload nodes were already configured and running and it got past that step and then went on to properly add the monitoring pods to the infra nodes:https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/default_storage_class/19/console

Do you think this is sufficient?

qiliRedHat:
@paigerube14 Yeah, I think it's good. That means we can rerun it without any manual work. @svetsa-rh What do you think?

svetsa-rh:
@paigerube14 @qiliRedHat 
I did not run into this particular one, but this looks good to me and will definitely help. 
Thanks for checking!

PR #269, opened 2022-09-23
Added upload mode to NOPE tool

nathan-weinberg:
Resolves [NETOBSERV-596](https://issues.redhat.com/browse/NETOBSERV-596)


nathan-weinberg:
//c @memodi @Amoghrd

nathan-weinberg:
Successfully tested locally:
```bash
[nathan@nathan-redhat scripts (netobserv-perf-tests)]$ ./nope.py --upload-file data_2022-09-21_14-40-05.json
INFO:root:Running NOPE tool in Upload mode - data from /home/nathan/ocp/ci/ocp-qe-perfscale-ci/data/data_2022-09-21_14-40-05.json will be uploaded to Elasticsearch
INFO:elasticsearch:POST https://search-ocp-qe-perf-scale-test-elk-hcm7wtsqpxy7xogbu72bor4uve.us-east-1.es.amazonaws.com:443/netobserv-perf/_doc [status:201 request:0.428s]
INFO:elasticsearch:POST https://search-ocp-qe-perf-scale-test-elk-hcm7wtsqpxy7xogbu72bor4uve.us-east-1.es.amazonaws.com:443/netobserv-perf/_doc [status:201 request:0.067s]
INFO:elasticsearch:POST https://search-ocp-qe-perf-scale-test-elk-hcm7wtsqpxy7xogbu72bor4uve.us-east-1.es.amazonaws.com:443/netobserv-perf/_doc [status:201 request:0.113s]
...
```

memodi:
@nathan-weinberg - I was thinking about this feature as sometimes uploads might fail transiently and by the time we get to it cluster is no longer available. 
I wonder, if capturing data in JSON prior to upload would be a good idea and having that archived as Jenkins artifacts so that we can upload locally even in cases where cluster dies later, WDYT?

nathan-weinberg:
> @nathan-weinberg - I was thinking about this feature as sometimes uploads might fail transiently and by the time we get to it cluster is no longer available. I wonder, if capturing data in JSON prior to upload would be a good idea and having that archived as Jenkins artifacts so that we can upload locally even in cases where cluster dies later, WDYT?

This already happens in Jenkins - if an upload fails the file is dumped to artifacts and can be downloaded/upload to ES manually

memodi:
> > @nathan-weinberg - I was thinking about this feature as sometimes uploads might fail transiently and by the time we get to it cluster is no longer available. I wonder, if capturing data in JSON prior to upload would be a good idea and having that archived as Jenkins artifacts so that we can upload locally even in cases where cluster dies later, WDYT?
> 
> This already happens in Jenkins - if an upload fails the file is dumped to artifacts and can be downloaded/upload to ES manually

Awesome, thanks!

scripts/nope.py
```diff
@@ -310,8 +310,8 @@ def main():
     # set prometheus flags
     parser.add_argument("--yaml-file", type=str, default='netobserv_queries_ebpf.yaml', help='YAML file from which to source Prometheus queries - defaults to "netobserv_queries_ebpf.yaml"')
     parser.add_argument("--user-workloads", default=False, action='store_true', help='Flag to query userWorkload metrics. Ensure FLP service and service-monitor are enabled and some network traffic exists.')
-    parser.add_argument("--starttime", type=str, required=True, help='Start time for range query')
-    parser.add_argument("--endtime", type=str, required=True, help='End time for range query')
+    parser.add_argument("--starttime", type=str, help='Start time for range query')
```
memodi, line 313: a better approach would be use argsparse [argument groups](https://docs.python.org/3/library/argparse.html#argument-groups), where you can have these has required arguments if in non-upload mode.

scripts/nope.py
```diff
@@ -332,6 +333,25 @@ def main():
     else:
         logging.basicConfig(level=logging.INFO)
 
+    # if running in upload mode - immediately load JSON, upload, and exit
+    UPLOAD_FILE = args.upload_file
+    if UPLOAD_FILE != '':
```
memodi, line 338:
```suggestion
    if UPLOAD_FILE is not None:
```

I am getting error as:
```
$ ./nope.py
Traceback (most recent call last):
  File "/Users/memodi/workspaces/repos/ocp-qe-perfscale-ci/scripts/./nope.py", line 339, in <module>
    upload_file_path = DATA_DIR + '/' + UPLOAD_FILE
TypeError: can only concatenate str (not "NoneType") to str
```

it would be also better to print `help` showing all arguments if no arguments are passed.

scripts/nope.py
```diff
@@ -342,9 +362,13 @@ def main():
     START_TIME = args.starttime
     END_TIME = args.endtime
     STEP = args.step
-    logging.info("Parsed Start Time: " + datetime.datetime.fromtimestamp(int(START_TIME)).strftime('%I:%M%p%Z UTC on %m/%d/%Y'))
-    logging.info("Parsed End Time:   " + datetime.datetime.fromtimestamp(int(END_TIME)).strftime('%I:%M%p%Z UTC on %m/%d/%Y'))
-    logging.info("Step is:           " + STEP)
+    if START_TIME == '' or END_TIME == '':
```
memodi, line 365: if you use argsparse arguments groups, you won't need to add such conditions, argsparse would handle it for you.

memodi (changes requested):
as CLI args grows suggest to use argsparse argument groups.

scripts/nope.py
nathan-weinberg: Ooh I hadn't heard of this, will check it out!

memodi: I use this in [jenkins_remote.py](https://github.com/memodi/tools/blob/master/jenkins_remote/jenkins_remote.py#L131) if you need reference

scripts/nope.py
nathan-weinberg: Actually I think we still need it, argparse doesn't seem to handle multiple exclusive groups well

scripts/nope.py
```diff
@@ -307,20 +307,21 @@ def main():
...
+    standard.add_argument("--endtime", type=str, help='End time for range query')
+    standard.add_argument("--step", type=str, default='10', help='Step time for range query')
+    standard.add_argument("--jenkins-job", type=str, help='Jenkins job name to associate with run')
+    standard.add_argument("--jenkins-build", type=str, help='Jenkins build number to associate with run')
+    standard.add_argument("--uuid", type=str, help='UUID to associate with run - if none is provided one will be generated')
+    standard.add_argument("--dump", default=False, action='store_true', help='Flag to dump data locally instead of uploading it to Elasticsearch')
+
+    # set upload mode flags
+    upload = parser.add_argument_group("Upload Mode", "Directly upload data from a previously generated JSON file to Elasticsearch")
+    upload.add_argument("--upload-file", type=str, default='', help='JSON file to upload to Elasticsearch. Must be in the "data" directory. Note this flag runs the NOPE tool in Upload mode and causes all flags other than --debug to be IGNORED.')
```
memodi, line 324: can we use the absolute user provided path instead of being very strict about where the UPLOAD_FILE should be located?

scripts/nope.py
```diff
@@ -332,6 +333,25 @@ def main():
     else:
         logging.basicConfig(level=logging.INFO)
 
+    # if running in upload mode - immediately load JSON, upload, and exit
+    UPLOAD_FILE = args.upload_file
+    if UPLOAD_FILE != '':
+        upload_file_path = DATA_DIR + '/' + UPLOAD_FILE
```
memodi, line 339: if we use user provided path then this won't be needed I think

memodi (changes requested):
some more comments

scripts/nope.py
```diff
@@ -342,9 +362,13 @@ def main():
     START_TIME = args.starttime
     END_TIME = args.endtime
     STEP = args.step
-    logging.info("Parsed Start Time: " + datetime.datetime.fromtimestamp(int(START_TIME)).strftime('%I:%M%p%Z UTC on %m/%d/%Y'))
-    logging.info("Parsed End Time:   " + datetime.datetime.fromtimestamp(int(END_TIME)).strftime('%I:%M%p%Z UTC on %m/%d/%Y'))
-    logging.info("Step is:           " + STEP)
+    if START_TIME == '' or END_TIME == '':
+        logging.error("START_TIME and END_TIME are needed to proceed")
+        sys.exit(1)
+    else:
```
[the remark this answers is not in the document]
nathan-weinberg: See my previous comments

scripts/nope.py
nathan-weinberg: Data should be saved in that path anyway, so IMO it makes sense to upload from there as well. We can always enhance this later.

scripts/nope.py
nathan-weinberg: See previous comments

scripts/nope.py
```diff
@@ -307,20 +307,21 @@ def main():
...
-    parser.add_argument("--uuid", type=str, help='UUID to associate with run - if none is provided one will be generated')
-
-    # set elasticsearch flags
-    parser.add_argument("--dump", default=False, action='store_true', help='Flag to dump data locally instead of uploading it to Elasticsearch')
+    # set standard mode flags
+    standard = parser.add_argument_group("Standard Mode", "Connect to an OCP cluster and gather data via Prometheus queries")
+    standard.add_argument("--yaml-file", type=str, default='netobserv_queries_ebpf.yaml', help='YAML file from which to source Prometheus queries - defaults to "netobserv_queries_ebpf.yaml"')
+    standard.add_argument("--user-workloads", default=False, action='store_true', help='Flag to query userWorkload metrics. Ensure FLP service and service-monitor are enabled and some network traffic exists.')
+    standard.add_argument("--starttime", type=str, help='Start time for range query')
+    standard.add_argument("--endtime", type=str, help='End time for range query')
```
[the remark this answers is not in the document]
memodi: right, they're not required if running with `--upload--file`, however when I am running in standard mode, I get this error if I don't specify those args, I run into this ugly exception:
```
$ ./nope.py
Traceback (most recent call last):
  File "/Users/memodi/workspaces/repos/ocp-qe-perfscale-ci/scripts/./nope.py", line 369, in <module>
    logging.info("Parsed Start Time: " + datetime.datetime.fromtimestamp(int(START_TIME)).strftime('%I:%M%p%Z UTC on %m/%d/%Y'))
TypeError: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```
since you're also checking for https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/259486569cf8f6025f80bb11bfbfd1369da1592a/scripts/nope.py#L365-L367 , having `--required=True` for this args makes sense and we can get rid of these lines to simplify code a bit.

> Actually I think we still need it, argparse doesn't seem to handle multiple exclusive groups well
could you share what issues you ran into?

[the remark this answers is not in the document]
nathan-weinberg: You can't have two exclusive argparse groups each with required arguments, and you can't use `required` with an argument if it's part of an exclusive group

As for the error, yes it's not pretty but tbh I'm not sure I see the value in handholding users for an internal tool that have not followed the README steps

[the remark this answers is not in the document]
memodi: okay, sounds good to me.

PR #275, opened 2022-09-28
Adding cerberus and cleanup to regression-test

paigerube14:
Adding in option to run cerberus and cleanup, when passing blank workload it'll clean up all namespaces that have the kube-burner-job label no matter the value

Needs to be added after PR: https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/274

https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/regression-test/6/console

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,6 +46,16 @@ pipeline {
                SOMEVARn='envn-test'<br>
                </p>'''
             )
+    booleanParam(
+      name: 'CLEANUP',
+      defaultValue: false,
+      description: 'Cleanup namespaces (and all sub-objects) created from workload (will run <a href=https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/benchmark-cleaner/>benchmark-cleaner</a>)'
+    )
+    booleanParam(
+      name: 'CERBERUS_CHECK',
+      defaultValue: false,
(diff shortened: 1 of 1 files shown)


qiliRedHat:
@paigerube14 I just realized I've had the cerberus check in this [job](https://gitlab.cee.redhat.com/aosqe/jenkins-jcasc-n/-/blob/master/jobs/scale-ci/perfscale-nightly-regression-longrun.jenkinsfile#L379) too. I will take out the cerberus check in that job to avoid duplicate checks. But before that do you think it's better to decouple the Regression job and the Ceberus job? In other jobs like kube-burner, router-perf we didn't include Ceberus job, but calling the cerberus job from a higher level jobs like scale-nightly-regression. The good thing of doing this is we can get seperate results of the kube-burner job and the cerberus job. If we put them together the higher level job had to read the 'final result' from the regression-test and phase it to know if it is a test case failure or ceberus check failure.

qiliRedHat:
@paigerube14 Some of my think in above comment is wrong. I just saw you have also added cerberus check in kube-burner in PR https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/277/files. So do you want to take out the cerberus check from here https://gitlab.cee.redhat.com/aosqe/jenkins-jcasc-n/-/blob/master/jobs/scale-ci/scale-nightly-regression.jenkinsfile#L393 too?

qiliRedHat:
@paigerube14 I understood now! In here https://gitlab.cee.redhat.com/aosqe/jenkins-jcasc-n/-/blob/master/jobs/scale-ci/scale-nightly-regression.jenkinsfile#L378 you didn't pass CERBERUS_CHECK=true and it is by default false.

PR #281, opened 2022-09-30
Updating metrics to generate

paigerube14:
With [PR](https://github.com/cloud-bulldozer/e2e-benchmarking/pull/411#event-7495808244) we decided to move to using container CPU and container memory because those metrics were already being captured. So we now need to update the csv generator with that data instead of pods

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -107,7 +107,7 @@ pipeline {
         )
         string(
             name: "COMPARISON_CONFIG",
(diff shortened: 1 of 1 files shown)


jtaleric (approved):
/lgtm

paigerube14:
Properly generating sheet: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/kube-burner/92/
NOTE: Failed because of pod metrics timing

PR #282, opened 2022-09-30
Changes to match updated ES schema

nathan-weinberg:
We need to make some ES schema changes for the long-lived Grafana dash - this PR ensures our tooling (specifically NOPE) will be compatible with those changes.

### DIFF
--- a/scripts/nope.py
+++ b/scripts/nope.py
@@ -63,6 +63,7 @@ def process_query(metric_name, query, raw_data):
     clean_data = []
 
     # capture metadata and value artifacts from prometheus return object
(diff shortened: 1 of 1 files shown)


scripts/nope.py
```diff
@@ -73,6 +73,7 @@ def process_query(metric_name, query, raw_data):
                         "uuid": UUID,
                         "metric_name": metric_name,
                         "data_type": "datapoint",
+                        "timestamp": data_point[0],
```
memodi, line 76: we'll need to have timestamp in milliseconds, simply need to multiply these by 1000 before upload.

PR #283, opened 2022-10-03
Added spreadsheet generation via workload jobs

nathan-weinberg:
Resolves [NETOBSERV-571](https://issues.redhat.com/browse/NETOBSERV-571)


paigerube14:
looks like you have an extra blank file here, want to remove that?

Jenkinsfile
```diff
@@ -403,23 +408,27 @@ pipeline {
                 script {
                     if (params.WORKLOAD == 'router-perf') {
                         env.JENKINS_JOB = 'scale-ci/e2e-benchmarking-multibranch-pipeline/router-perf'
+                        COMPARISON_CONFIG_FILES = 'clusterVersion.json podLatency.json podCPU-avg.json podCPU-max.json podMemory-avg.json podMemory-max.json netobserv-touchstone.json'
```
paigerube14, line 411: the data for the router perf job is a little different so these files aren't reusable in this workload 

The data for this workload gets stored in pattern "router-test-results" as you can see in this file. https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/router-perf-v2/mb-touchstone.json

paigerube14: multiple files yes, but you'll just want 'mb-touchstone.json netobserv-touchstone.json'

Jenkinsfile
```diff
@@ -569,21 +579,24 @@ pipeline {
...
+                        env.WORKLOAD = params.WORKLOAD
                         returnCode = sh(returnStatus: true, script: """
                             python3.9 --version
                             python3.9 -m pip install virtualenv
                             python3.9 -m virtualenv venv3
                             source venv3/bin/activate
                             python --version
-                            cd $WORKSPACE/benchmark-comparison
-                            python setup.py develop
-                            export UUID=`jq -r '.uuid' $WORKSPACE/ocp-qe-perfscale-ci/data/workload.json`
```
paigerube14, line 585: You are still going to want this UUID. You'll also need NETWORK_TYPE and to set a WORKLOAD just for file naming

Jenkinsfile
```diff
@@ -569,21 +579,24 @@ pipeline {
...
+                    userRemoteConfigs: [[url: 'https://github.com/cloud-bulldozer/e2e-benchmarking.git' ]],
+                    extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'e2e-benchmarking']]
                 ])
-                withCredentials([usernamePassword(credentialsId: 'elasticsearch-perfscale-ocp-qe', usernameVariable: 'ES_USERNAME', passwordVariable: 'ES_PASSWORD')]) {
+                withCredentials([usernamePassword(credentialsId: 'elasticsearch-perfscale-ocp-qe', usernameVariable: 'ES_USERNAME', passwordVariable: 'ES_PASSWORD'), file(credentialsId: 'sa-google-sheet', variable: 'GSHEET_KEY_LOCATION')]) {
                     script {
+                        env.COMPARISON_CONFIG = COMPARISON_CONFIG
+                        env.ES_SERVER = "https://$ES_USERNAME:$ES_PASSWORD@search-ocp-qe-perf-scale-test-elk-hcm7wtsqpxy7xogbu72bor4uve.us-east-1.es.amazonaws.com"
+                        env.EMAIL_ID_FOR_RESULTS_SHEET = "${userId}@redhat.com"
+                        env.GEN_CSV = params.GEN_CSV
```
paigerube14, line 590: You shouldn't need to set the parameters to environment variables, it should just use them (if they have the same name). I don't outright set the WORKLOAD environment variable and running kube burner works fine here: https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/kube-burner/Jenkinsfile

Just in case you could do this, 
```
 echo "$ENV_VARS" > .env_override
# Export those env vars so they could be used by CI Job
 set -a && source .env_override && set +a
```

Jenkinsfile
paigerube14: NETWORK TYPE you can set from: $(oc get network.config/cluster -o jsonpath='{.spec.networkType}')

Jenkinsfile
nathan-weinberg: [You are right](https://stackoverflow.com/questions/50398334/what-is-the-relationship-between-environment-and-parameters-in-jenkinsfile-param), but I may do this just to keep everything straight :laughing: we aren't using the ENV_VARS block yet, but if we introduce it I will add this in

PR #296, opened 2022-10-11
Benchmark cleaner with no benchmark operator

paigerube14:
E2e benchmarking has removed the usage of benchmark operator so some of the items that we try to delete in this job no longer exist. Taking out the bad references


paigerube14:
Run here: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/benchmark-cleaner/22/console

PR #308, opened 2022-10-27
image tag fix and more queries

memodi:
also, removing group_left queries since its not playing well over http api. I'll do  PR for e2e-benchmarking to update touchstone config.


nathan-weinberg (commented):
/lgtm

PR #314, opened 2022-11-04
Not properly setting status if cerberus fails

paigerube14:
If cerberus fails it is setting an incorrect variable

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -143,11 +143,14 @@ pipeline {
               export folder=${SCRIPT%/*}
               export script=${SCRIPT##*/} 
               cd ${folder}
+
+              chmod +x ${script}
               set -o pipefail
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -162,14 +165,16 @@ pipeline {
                       string(name: 'JENKINS_AGENT_LABEL', value: JENKINS_AGENT_LABEL),booleanParam(name: "INSPECT_COMPONENTS", value: true)
                   ],
                   propagate: false
-              if (status == "PASS") {
-                  if (cerberus_job == null && cerberus_job == "" && cerberus_job.result.toString() != "SUCCESS") {
-                      status = "Cerberus check failed"
+              if (RETURNSTATUS == "PASS") {
```
qiliRedHat, line 168: @paigerube14 def RETURNSTATUS = "default" and nowhere has set RETURNSTATUS to"PASS"

PR #320, opened 2022-11-17
updates to dashboard, queries, loki CRD

memodi:
Also, increasing the default size in nope.py since I don't think we need 10s granular data.


memodi:
@nathan-weinberg - `replicationFactor : 1` is probably why we were able to install small/medium LokiStack on smaller size machines. Now that it's removed it will use default `replicationFactor` which are below and likely will only fit for machine sizes recommended on Jenkins jobs:
```
1 - 1x.extra-small
2 - 1x.small
3 - 1x.medium
```

PR #327, opened 2022-12-08
Adding parallel runs of kraken,cerberus,upgrade(if set)

paigerube14:
Adding job to be able to run cerberus, kraken and an upgrade in parallel 

https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/chaos_upgrade/41/console 
(failure only because I had to abort Cerberus b/c I gave too many iterations) 

Currently have to hard code number of iterations for cerberus to run. Each iteration (with the 3 second wait) takes between 7-10 seconds and an upgrade takes any where from 1.2 -3 hours. 
Note: For an upgrade that latest 1.2 hours I ran about 500 iterations in the same time

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,260 @@
+@Library('flexy') _
+
+// rename build
+def userId = currentBuild.rawBuild.getCause(hudson.model.Cause$UserIdCause)?.userId
+if (userId) {
+  currentBuild.displayName = userId
+}
+
+def RETURNSTATUS = "default"
+def output = ""
+
+def kraken_job = ""
+def cerberus_job = ""
+def upgrade_ci = ""
+
+pipeline {
+  agent none
+  parameters {
+        string(
+          name: 'BUILD_NUMBER', 
+          defaultValue: '', 
+          description: 'Build number of job that has installed the cluster.'
+        )
+        string(
+          name: 'UPGRADE_VERSION', 
+          description: 'This variable sets the version number you want to upgrade your OpenShift cluster to.'
+        )
+        booleanParam(
+          name: 'ENABLE_FORCE', 
+          defaultValue: true, 
+          description: 'This variable will force the upgrade or not'
+        )
+        string(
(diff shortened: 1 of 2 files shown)


qiliRedHat:
@paigerube14 Besides upgrade test, running cerberus and/or kraken in parallel with some other jobs could be a general function. For example, running scale-ci with cerberus + kraken, running some regression test with cerberus + kraken. Now this job needs manually setting the number of cerberus and cerberus iterations, so it's loose coupled with the upgrade job. Shall we make cerberus + kraken (can choose to run a single one or both) a separate job, and other jobs like scale-ci, upgrade, regression can be added a checkbox and some parameters to call it in a stage and run in parallel with the origin stages.

paigerube14:
> @paigerube14 Besides upgrade test, running cerberus and/or kraken in parallel with some other jobs could be a general function. For example, running scale-ci with cerberus + kraken, running some regression test with cerberus + kraken. Now this job needs manually setting the number of cerberus and kraken iterations, so it's loose coupled with the upgrade job. Shall we make cerberus + kraken (can choose to run a single one or both) a separate job, and other jobs like scale-ci, upgrade, regression can be added a checkbox and some parameters to call it in a stage and run in parallel with the origin stages.

Yes think that this type of functionality with calling all jobs in parallel can be added in the future but I was currently just trying to finish up work for the [chaos okr](https://issues.redhat.com/browse/OCPQE-12678) by end of year

There is going to have to be a lot of additional work to this job to make it work like loaded-upgrade (with all possible sub jobs and variables). 

Were you thinking that this job would have all the parallel calls or should we add a parallel option to each of the sub jobs you mentioned (kube-burner, regression, and upgrade)

qiliRedHat:
@paigerube14 Got you! Let's merge this. The rest can be considered step by step next year.

qiliRedHat (approved):

PR #354, opened 2023-02-07
Adding post result to slack

paigerube14:
Adding post result when job is finished


paigerube14:
Also adding call to must gather under cerberus if failure

PR https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/360 must go in first

PR #375, opened 2023-02-22
Adding churn option for kube-burner at top level

paigerube14:
After pr: https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/374 we will want to add this option at top level


paigerube14:
@skordas can you PTAL

PR #376, opened 2023-02-23
Adding NetworkPolicy test case

skordas:
Ticket: https://issues.redhat.com/browse/OCPQE-13658

PR #377, opened 2023-02-23
Adding Networkpolicy TC description.

skordas:
To be consistent with https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/376


skordas:
@paigerube14 PTAL

PR #382, opened 2023-03-02
adding churn and cluster density numbers increase

paigerube14:
Going along with pr https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/381 want to add churn on by default


paigerube14:
@svetsa-rh can you review this one as well since you approved the first one

PR #389, opened 2023-03-10
Added IIB override option to Internal installation

nathan-weinberg:
Based on Slack huddle

### DIFF
--- a/scripts/catalogsources/qe-unreleased-catalogsource.yaml
+++ b/scripts/catalogsources/qe-unreleased-catalogsource.yaml
@@ -5,5 +5,5 @@ metadata:
   namespace: openshift-marketplace
 spec:
   displayName: NetObservUnreleasedTesting
-  image: 'quay.io/openshift-qe-optional-operators/aosqe-index:v${MAJOR_VERSION}.${MINOR_VERSION}'
+  image: ${IMAGE}
   sourceType: grpc
--- a/scripts/netobserv.sh
+++ b/scripts/netobserv.sh
@@ -103,7 +103,7 @@ deploy_lokistack() {
   elif [[ $LOKISTACK_SIZE == "1x.medium" ]]; then
     LokiStack_CONFIG=$SCRIPTS_DIR/loki/lokistack-1x-medium.yaml
   else
-    echo "====> No LokiStack config was found - using 1x-exsmall"
+    echo "====> No LokiStack config was found - using 1x.extra-small"
     echo "====> To set config, set LOKISTACK_SIZE variable to either '1x.extra-small', '1x.small', or '1x.medium'"
     LokiStack_CONFIG=$SCRIPTS_DIR/loki/lokistack-1x-exsmall.yaml
   fi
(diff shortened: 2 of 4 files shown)


nathan-weinberg:
Test using `Internal` with `IIB_OVERRIDE`: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/nweinber-e2e-benchmarking-multibranch-pipeline/job/netobserv-perf-tests/53/
Test using `Internal` without `IIB_OVERRIDE`: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/nweinber-e2e-benchmarking-multibranch-pipeline/job/netobserv-perf-tests/54/

scripts/netobserv.sh
```diff
@@ -80,7 +80,7 @@ deploy_lokistack() {
   if [[ $WORKLOAD == "None" ]] || [[ -z $WORKLOAD ]]; then
     export S3_BUCKETNAME="netobserv-ocpqe-perf-loki-$RAND_SUFFIX"
   else
-    export S3_BUCKETNAME="netobserv-ocpqe-perf-loki-$WORKLOAD"
+    export S3_BUCKETNAME="netobserv-ocpqe-perf-loki-$WORKLOAD-$RAND_SUFFIX"
```
memodi, line 83: I guess, it's make sense if those old buckets does exist for runs made in succession, with this we'd be sure to use the new bucket every time - was that the reason to change this or there's more to it?

scripts/netobserv.sh
```diff
@@ -124,6 +124,11 @@ deploy_unreleased_catalogsource() {
   oc apply -f $SCRIPTS_DIR/icsp.yaml
 
   echo "====> Determining CatalogSource config"
+  if [[ -z $IMAGE ]]; then
```
memodi, line 127: Could be additional log with `echo` here to specify which image it ends up using in build log? 

I know we don't mention about installing from "Internal" on our Readme, just Source and OperatorHub. Do you think we should ? 
 if we do, we should also mention about setting `IMAGE` ,  `MAJOR_VERSION` and `MINOR_VERSION` .

memodi (changes requested):
looks good overall, just a minor comment/question.

scripts/catalogsources/qe-unreleased-catalogsource.yaml
```diff
@@ -5,5 +5,5 @@ metadata:
   namespace: openshift-marketplace
 spec:
   displayName: NetObservUnreleasedTesting
-  image: 'quay.io/openshift-qe-optional-operators/aosqe-index:v${MAJOR_VERSION}.${MINOR_VERSION}'
+  image: ${IMAGE}
```
Amoghrd, line 8: This is what I was asking about, I like this approach setting values in YAML files

scripts/netobserv.sh
nathan-weinberg: Honestly, the reason behind it was that I've seen collisions where more than one person was running the same workload or cases when a bucket wasn't cleaned up and there were errors due to the bucket already exisiting

scripts/netobserv.sh
nathan-weinberg: I will add an `echo` and update the `README`!

scripts/catalogsources/qe-unreleased-catalogsource.yaml
nathan-weinberg: Yes I am going to see if there is additional consolidation we can do in future PRs!

PR #390, opened 2023-03-14
Taking out force parameter

paigerube14:
Updates for both loaded-upgrade and kube-burner jobs needed

passing jobs of each type: 
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/write-to_sheet/79/parameters/

https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/write-to_sheet/85/console

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,7 +102,7 @@ pipeline {
             
             export PYTHONIOENCODING=utf8
             if [[ "${params.JOB}" == "loaded-upgrade" ]]; then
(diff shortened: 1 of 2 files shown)


svetsa-rh:
@paigerube14 Is there any reason why we don't want to track while writing results that ENABLE_FORCE was specified during  upgrade/loaded upgrade?
ENABLE_FORCE itself will still be available, so why not track if someone used it?

paigerube14:
> @paigerube14 Is there any reason why we don't want to track while writing results that ENABLE_FORCE was specified during upgrade/loaded upgrade? ENABLE_FORCE itself will still be available, so why not track if someone used it?

I don't currently track it on the loaded-upgrade sheet so I don't need to pass the variable 
It could be cross referenced on the upgrade sheet with the upgrade variable

svetsa-rh:
Hmm.. 
Does this mean we have to look at 2 different places while checking results? Typically, I see people only mentioning results from loaded upgrade in their jira tasks. 
Should they be specifying results from both loaded upgrade and upgrade tabs?

paigerube14:
@svetsa-rh yes. Also for completeness would probably also add the kube-burner job information as well. I tried to take out duplicate data on the loaded-upgrade job sheet and give links to the sub jobs that were run and give an overview of what was ran

svetsa-rh:
Super. 
/lgtm

PR #392, opened 2023-03-22
add infra, workload MCPs

jianzhangbjz:
After removing the default `worker` role from the `infra, workload` node, we should add the related `MCP` for them. Otherwise, it will lead to some configs cannot be applied. Details: https://issues.redhat.com/browse/OCPBUGS-10596


jianzhangbjz:
Hi @qiliRedHat , could you help approve it? Thanks!

PR #400, opened 2023-04-06
Adding blank option

paigerube14:
Need to add blank option to only run upgrade with chaos and cerberus

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,7 +48,7 @@ pipeline {
 			"""
         )
         choice(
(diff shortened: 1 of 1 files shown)


paigerube14:
In testing, I found that the upgrade would always run twice even if no test is selected: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/chaos-upgrade/2/console

skordas (approved):
/lgtm

PR #401, opened 2023-04-06
Added additional features to pipeline, updated CRDs (round 2)

nathan-weinberg:
- Exposed `eBPF` memory limit and `router-perf` workload options for configuration
- Added `DESTROY_CLUSTER` option to cleanup stage
- Added `PROFILE_PORT` to flowcollector CRD
- Minor updates to upstream Dittybopper config
- Reduced number of logs generated by FLP
- Extended NOPE tool to attempt multiple token capture methods
- Altered AWS deletion func to try and make it more reliable

### DIFF
--- a/scripts/netobserv.sh
+++ b/scripts/netobserv.sh
@@ -59,7 +59,7 @@ deploy_netobserv() {
     sleep 1
   done
   sleep 60
-  oc wait --timeout=180s --for=condition=ready pod -l app=flowlogs-pipeline -n netobserv
+  oc wait --timeout=600s --for=condition=ready pod -l app=flowlogs-pipeline -n netobserv
 }
 
 deploy_lokistack() {
@@ -235,13 +235,15 @@ setup_dittybopper_template() {
 }
 
 delete_s3() {
-  echo "====> Deleting AWS S3 Bucket"
   echo "====> Getting S3 Bucket Name"
   S3_BUCKET_NAME=$(/bin/bash -c 'oc extract cm/lokistack-config -n netobserv --keys=config.yaml --confirm --to=/tmp | xargs -I {} egrep bucketnames {} | cut -d: -f 2 | xargs echo -n')
   echo "====> Got $S3_BUCKET_NAME"
-  aws s3 rm s3://$S3_BUCKET_NAME --recursive
-  sleep 30
-  aws s3 rb s3://$S3_BUCKET_NAME --force
+  echo "====> Deleting AWS S3 Bucket"
+  while :; do
+    aws s3 rb s3://$S3_BUCKET_NAME --force && break
+    sleep 1
+  done
+  echo "====> AWS S3 Bucket $S3_BUCKET_NAME deleted"
 }
 
 delete_lokistack() {
--- a/scripts/nope.py
+++ b/scripts/nope.py
@@ -464,8 +464,14 @@ if __name__ == '__main__':
     IS_DOWNSTREAM = subprocess.run(['oc', 'get', 'pods', '-l', 'app=netobserv-operator', '-o', 'jsonpath="{.items[*].spec.containers[0].env[-2].value}"', '-A'], capture_output=True, text=True).stdout
     if IS_DOWNSTREAM == '"true"':
         TOKEN = subprocess.run(['oc', 'create', 'token', 'prometheus-k8s', '-n', 'openshift-monitoring'], capture_output=True, text=True).stdout
+        # try deprecated method in case first attempt fails
+        if TOKEN == '':
+            TOKEN = subprocess.run(['oc', 'sa', 'new-token', 'prometheus-k8s', '-n', 'openshift-monitoring'], capture_output=True, text=True).stdout
     else:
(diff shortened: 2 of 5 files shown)


scripts/loki/lokistack-1x-small.yaml
```diff
@@ -8,7 +8,7 @@ spec:
     global:
       ingestion:
         ingestionBurstSize: 100
-        ingestionRate: 500
+        ingestionRate: 5000
```
memodi, line 11: are we looking to change 1x-small? asking because I know we use 1x-medium mostly for Perf testing

scripts/nope.py
```diff
@@ -464,8 +464,14 @@ def main():
     IS_DOWNSTREAM = subprocess.run(['oc', 'get', 'pods', '-l', 'app=netobserv-operator', '-o', 'jsonpath="{.items[*].spec.containers[0].env[-2].value}"', '-A'], capture_output=True, text=True).stdout
     if IS_DOWNSTREAM == '"true"':
         TOKEN = subprocess.run(['oc', 'create', 'token', 'prometheus-k8s', '-n', 'openshift-monitoring'], capture_output=True, text=True).stdout
+        # try deprecated method in case first attempt fails
+        if TOKEN == '':
```
memodi, line 468: 😮  curious, in what cases new method fails?

memodi (commented):
@nathan-weinberg - looks good, just couple of minor questions.

scripts/nope.py
Amoghrd: yeah this one is seen on a 4.10 cluster. It uses a deprecated version of create token which causes NOPE to fail. Nathan probably can might give more insight tho

scripts/loki/lokistack-1x-small.yaml
nathan-weinberg: @memodi yes, in discussions with Steven and Sara we found some of our test cases could use `1x-small` instead of `1x-medium` so looking to use that where we can due to lower resource requirements on the cluster nodes

as for this ingestion rate, still playing with this, I was trying to mitigate the problem we faced in [NETOBSERV-975](https://issues.redhat.com/browse/NETOBSERV-975)

scripts/nope.py
nathan-weinberg: exactly, i believe this is due to differences between `oc` binary on `oc410` agents - since it is an older version, the `oc create` command was not creating a token during 4.10 perf runs (we had to use 4.10 for a period due to [OCPBUGS-11636](https://issues.redhat.com/browse/OCPBUGS-11636)) causing the NOPE tool to fail. my hope is this will mitigate that :smile:

nathan-weinberg:
@memodi thanks, some more tested needed for this but @Amoghrd and I have been using my forked job for much of the last couple weeks so fairly confident about most of it :joy:

scripts/loki/lokistack-1x-small.yaml
nathan-weinberg: @memodi @Amoghrd since this change is a WIP I'm thinking of saving it for a followup PR so we can merge this one, thoughts?

scripts/netobserv.sh
```diff
@@ -59,7 +59,7 @@ deploy_netobserv() {
     sleep 1
   done
   sleep 60
-  oc wait --timeout=180s --for=condition=ready pod -l app=flowlogs-pipeline -n netobserv
+  oc wait --timeout=600s --for=condition=ready pod -l app=flowlogs-pipeline -n netobserv
```
nathan-weinberg, line 62: @memodi btw, if you were curious why I did this, on large clusters (e.g. 120 nodes) FLP pods can take 5-10min to come up. not sure if that's something we might consider a bug but good to have increased timeout here so we don't have extra pipeline failures (once they come up everything is fine)

scripts/netobserv.sh
```diff
@@ -235,13 +235,15 @@ setup_dittybopper_template() {
...
   echo "====> Got $S3_BUCKET_NAME"
-  aws s3 rm s3://$S3_BUCKET_NAME --recursive
-  sleep 30
-  aws s3 rb s3://$S3_BUCKET_NAME --force
+  echo "====> Deleting AWS S3 Bucket"
+  while :; do
+    aws s3 rb s3://$S3_BUCKET_NAME --force && break
+    sleep 1
+  done
+  echo "====> AWS S3 Bucket $S3_BUCKET_NAME deleted"
```
nathan-weinberg, line 246: @memodi curious what you think of this change, I based it off some code you wrote :smile: worked for @Amoghrd [in this job](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/nweinber-e2e-benchmarking-multibranch-pipeline/job/netobserv-perf-tests/225/parameters/)

scripts/loki/lokistack-1x-small.yaml
memodi: Sure, no problem.

scripts/netobserv.sh
nathan-weinberg: Unfortunately, yes :sob: esp when the bucket has lots of objects in it. think it's a bug with the `aws` cli - but for our purposes, this should handle all cases

memodi (approved):
/lgtm

PR #408, opened 2023-04-20
Using output file from artifacts and sandman

paigerube14:
Using Nathans script of [sandman](https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/netobserv-perf-tests/scripts/sandman.py) to parse output files and be able to more easily find start/end times as well as uuid

This needs to go in at the same time as all other workloads to be able to properly pass the new parameters properly. Will update other workloads soon


write_to_sheet/kb.out
```diff
@@ -0,0 +1,25 @@
+###############################################
```
qiliRedHat, line 1: is this file supposed to be added? Looks like a test output.

PR #451, opened 2023-06-13
Adding functionality for automatic comparsion using Tolerancy Rules

nathan-weinberg:
Resolves [NETOBSERV-1095](https://issues.redhat.com/browse/NETOBSERV-1095)


memodi (commented):
thanks @nathan-weinberg . As discussed in sync adding json output of tolerancy and saving it as artifact is only pending here, can be done in follow up PR.

/lgtm

PR #452, opened 2023-06-14
touchstone config for tolerancy

memodi:
NETOBSERV-1094 Need a separate touchstone_config.json for tolerancy/comparison with only metrics that we look to compare.


Amoghrd:
Think new line was required at EOF

memodi:
> Think new line was required at EOF

yeah, I tried with vscode but it reverts it back. I guess it should be okay as it won't hurt the functionality.

PR #455, opened 2023-06-15
Adding more workload type and get data for network-perf type

paigerube14:
Missing more workload types that are available in kube-burner and kube-burner-ocp

Want to also be able to get the workload type for network perf in the results post to es


paigerube14:
@qiliRedHat PTAL

PR #461, opened 2023-06-29
Post results to es

paigerube14:
Want to see if the cluster has fips installed on it and if its across multiple zones


mffiedler:
Nice additions

PR #462, opened 2023-06-29
Kube burner remove workloads

paigerube14:
Removing max-namespace and max-services


svetsa-rh:
@paigerube14 Didn't @rpattath say not to make these changes before 4.14 release? May be I misunderstood. You may want to confirm with her.

PR #465, opened 2023-06-29
Moved Mr. Sandman from netobserv to main branch

nathan-weinberg:
This script, formerly only used by NetObserv QE, is now being used by PerfScale QE as well. As such, @paigerube14 and I discussed moving it to the `main` branch to make collaboration easier.


scripts/requirements.txt
```diff
@@ -0,0 +1 @@
+argparse>=1.4.0
```
paigerube14, line 1: think we also need datetime

paigerube14: nvm, ran without needing to install

paigerube14 (approved):
/lgtm

paigerube14:
@nathan-weinberg not sure why cluster density v2 on rosa is a required test here. Need to look into that. think it'll block the merging...

nathan-weinberg:
@paigerube14 @svetsa-rh no worries! no rush on this - let me know if you need anything from my side!

svetsa-rh:
@nathan-weinberg No action needed from your side for now. @paigerube14 is going to remove one of the files and merge changes. If the above test fails you can go ahead and rebase and pull her changes to skip this test.

paigerube14:
The one enhancement I had mentioned to Nathan is to add a bash file that has all of the items in the dict set up as environment variables so we'll be able to export the file and not need to parse out every variable to set it. Will be a follow up PR for that

PR #470, opened 2023-07-05
Making sourcable file

paigerube14:
Want to be able to add a .sh file that we can export in our scripts instead of jq every variable every time 

Also adding a try except for if an error occurs inthe run and the end ouptut is not found for some reason, to find the last time regex (bug for reference: https://issues.redhat.com/browse/OCPQE-15860) CC: @SachinNinganure

### DIFF
--- a/scripts/sandman.py
+++ b/scripts/sandman.py
@@ -82,8 +82,15 @@ def main():
     
     # capture and log strings representations of start and end times
     starttime_string = re.findall(starttime_regex, workload_logs)[0]
-    endtime_string = re.findall(endtime_regex, workload_logs)[0]
+    
     print(f"starttime_string: {starttime_string}")
+
+    try: 
+        endtime_string = re.findall(endtime_regex, workload_logs)[0]
+    except: 
+        # if can't find the end time properly (error during run)
+        # find the last time posted in workload_logs file
+        endtime_string = re.findall(base_regex, workload_logs)[-1]
     print(f"endtime_string: {endtime_string}")
 
     # convert string times to unix timestamps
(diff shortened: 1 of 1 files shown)


nathan-weinberg:
Will try this with NetObserv, please DNM until then

/hold

scripts/sandman.py
```diff
@@ -123,6 +130,13 @@ def main():
     with open(DATA_DIR + f'/workload.json', 'w') as data_file:
         json.dump(workload_data, data_file)
 
+    workload_env_vars = ""
+    for k,v in workload_data.items():
+        workload_env_vars += "export " + k + "='" + v + "'\n"
+    # write timestamp data to data directory
+    with open(DATA_DIR + f'/workload.sh', 'w') as data_file:
+        data_file.write(workload_env_vars)
```
nathan-weinberg, line 138: do you want to wrap this in a `try/catch` statement in case something fails with the iteration? JSON file could then be used for debugging

scripts/sandman.py
```diff
@@ -123,6 +130,13 @@ def main():
     with open(DATA_DIR + f'/workload.json', 'w') as data_file:
         json.dump(workload_data, data_file)
 
+    workload_env_vars = ""
+    for k,v in workload_data.items():
+        workload_env_vars += "export " + k + "='" + v + "'\n"
+    # write timestamp data to data directory
```
nathan-weinberg, line 136: probably don't need to duplicate this comment, I'd maybe update the above one to be more descriptive as to what is going on here now (creating a JSON as well as a Shell file)

scripts/sandman.py
paigerube14: The json will still be printed already right?

nathan-weinberg: Yes, and come to think of it we will want this to fail if it doesn't work, so we can probably leave it as is and enhance later

scripts/sandman.py
nathan-weinberg: I reckon at some point in the future we could even remove the JSON generation, but let's leave it for now

paigerube14: yep! will leave but update comment

PR #474, opened 2023-07-06
Adding buildInfo references

paigerube14:
Can easily set env variables from build json


nathan-weinberg:
two minor comments:
- might want to move `UUID` comment to right above where it is set
- do we want to call it `buildInfo`? might want to do something like `workloadInfo` or something else

anyway both are nits, lmk your thoughts and I can approve either way :+1:

paigerube14:
@nathan-weinberg updated based on your comments, ptal

PR #475, opened 2023-07-06
More efficient parsing of Mr. Sandman output file

nathan-weinberg:
Based off convos with @paigerube14 

Also did some minor cleanup across our stack

PR #476, opened 2023-07-06
Made Sandman output configurable

nathan-weinberg:
Building off of https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/470 this PR does the following:
- adds a `.gitignore` for Sandman-related artifacts
- adds in the `README.md` blurb being removed in https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/475
- allows users to config which output file type they wish to specify between `json` and `sh` - since currently teams are only using JSON, this field is optional and defaults to JSON, but we may want to explore adding additional file types as well such as YAML sometime in the future


nathan-weinberg:
I tried out a more holistic update for the `main` branch `README` - lmk your thoughts @mffiedler @paigerube14 - can always revert to the previous commit

scripts/sandman.py
```diff
@@ -126,31 +127,36 @@ def main():
...
 if __name__ == '__main__':
 
     # initialize argument parser
-    parser = argparse.ArgumentParser(description='Mr. Sandman')
+    parser = argparse.ArgumentParser(description='Mr. Sandman: Master of Time and the Knowledge bound by it')
 
     # set argument flags
-    parser.add_argument("--file", type=str, required=True, help='File to parse')
+    parser.add_argument("--file", type=str, required=True, help='Workload out file to parse')
+    parser.add_argument("--output", type=str, default='json', help="Sandman out file type - options are 'json' (default) and 'sh'")
```
memodi, line 151: looking at the description of the arg, I don't know what 'sh' is? I can figure out if I looked in code, it will write bunch of variables to sh script but won't be able to understand if I didn't know code.  Is there a need for multiple file format? if there's really a need for both then perhaps we should explain it better.

scripts/sandman.py
```diff
@@ -126,31 +127,36 @@ def main():
...
 
     # parse arguments
     args = parser.parse_args()
     WORKLOAD_OUT_FILE = args.file
+    SANDMAN_OUT_FILE_TYPE = args.output
+    if SANDMAN_OUT_FILE_TYPE not in ['json', 'sh']:
```
memodi, line 157: within parse.add_argument you can have `choices`, argsparse will throw exception if its outside of those choices and no need for `if` condition

memodi (changes requested):
just couple of comments,

scripts/sandman.py
nathan-weinberg: @memodi that functionality was already added by @paigerube14, this just makes it into an option

scripts/sandman.py
nathan-weinberg: nice tip!

README.md
```diff
@@ -1,24 +1,57 @@
...
+| post-to-slack | TODO |
+| regression-test | TODO |
+| router-perf | TODO |
+| upgrade | TODO |
+| write-scale-ci-results | TODO |
+
+## Scripts
+
+### Mr. Sandman
+Sometimes, the only way to get data such as UUID and workload timestamp information is directly from the workload job runs. If you find yourself in need of this but don't want to manually pour through logs, you can let Mr. Sandman give it a shot by running `./scripts/sandman.py --file <path/to/out/file>`
```
paigerube14, line 50: Want to add file type option here

nathan-weinberg: Done!

scripts/sandman.py
nathan-weinberg: Clarified in the `README`

paigerube14:
looks good!

PR #484, opened 2023-07-17
update prow ci add infra/workload and move pods script to replace old script

liqcui:
To support vsphere8,  we can use the same script of prow ci to add infra/workload and move monitoring/registry/ingress pods script. 

We don't remove old script, just adding new script. 

The Jenkins job is https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/liqcui-e2e-benchmarking-multibranch-pipeline/job/cluster-post-config/195/console

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -141,7 +141,8 @@ pipeline {
               set -x
 
               SECONDS=0
-              ./install_infra_work.sh
+              ./openshift-qe-workers-infra-workload-commands.sh
+              ./openshift-qe-move-pods-infra-commands.sh
               status=$?
               echo "final status $status"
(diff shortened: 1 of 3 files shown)


paigerube14:
@liqcui once we add these scripts it would be good if we could maybe update the prow job to clone this repo/branch code. So we would be able to know it has the same code and don't get prow and jenkins out of sync.

liqcui:
> @liqcui once we add these scripts it would be good if we could maybe update the prow job to clone this repo/branch code. So we would be able to know it has the same code and don't get prow and jenkins out of sync.

I will update prow ci later and make it consistency

PR #491, opened 2023-07-31
Separated code for rosa upgrade

liqcui:
Description:
This PR move all function in upgrade.sh to common.sh, no any code changed, due to rosa upgrade is totally different than regular OCP, so can not directly use upgrade.sh, but we can re-use the function of checking upgrade status.


liqcui:
@paigerube14 @qiliRedHat Please review my PR when you are available, thanks!

upgrade_scripts/__pycache__/check_upgrade.cpython-36.pyc
qiliRedHat, line 1: Is file needed?

upgrade_scripts/check-rosa-upgrade.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+source ./common.sh
+upgrade_pass=True
+
+capture_failed_pods_before_upgrade
```
qiliRedHat, line 5: @liqcui In prow this script is executed in rosa-upgrade-postcheck, which is after the rosa-loaded-upgrade execution, right? So this is not actually 'before upgrade'?

upgrade_scripts/__pycache__/check_upgrade.cpython-36.pyc
liqcui: No need, it should be a compile file of python, let me delete it.

upgrade_scripts/check-rosa-upgrade.sh
liqcui: Yes, we can remove this, it's not suitable for rosa upgrade

liqcui: fixed in new code, thanks!

upgrade_scripts/__pycache__/check_upgrade.cpython-36.pyc
liqcui: fixed in new code, thanks!

upgrade_scripts/common.sh
```diff
@@ -0,0 +1,190 @@
...
+
+function capture_failed_pods_before_upgrade(){
+  #Save the failed job before upgrade and make sure new failed job caused by upgrade,ignore installer error
+  echo "Capture failed pods before upgrade OCP"
+  echo "####################################################################################"
+  oc get pods -A| grep -v -E 'Running|Completed|installer'| awk '(NF=NF-2) 1'>/tmp/upgrade-before-failed-pods.txt
+}
+
+function capture_failed_pods_after_upgrade(){
+  #Save the failed job before upgrade and make sure new failed job caused by upgrade
```
qiliRedHat, line 153:
```suggestion
  #Save the failed job after upgrade and make sure new failed job caused by upgrade
```

liqcui: @qiliRedHat capture_failed_pods_after_upgrade() relay on capture_failed_pods_before_upgrade, the capture_failed_pods_before_upgrade will capture all failed pods status, then save it to /tmp/failed-job-before-upgrade.txt, after the upgrade complete, will capture all failed pods status, save it to /tmp/failed-job-after-upgrade.txt  then compare the two files, if new failed pods exist, it will throw error, we have to add both or remove both.

liqcui: Sorry mis-understood what your meaning, I accept your suggested commit, the comments that describe wrong, it should [after], not [before]

upgrade_scripts/check-rosa-upgrade.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+source ./common.sh
+upgrade_pass=True
+
+SECONDS=0
+export PYTHONUNBUFFERED=1
+python3 -c "import check_upgrade; check_upgrade.check_upgrade('$target_version_prefix')"
+duration=$SECONDS
+echo "$(($duration / 60)) minutes and $(($duration % 60)) seconds elapsed."
+sleep 120
```
paigerube14, line 10: Is this sleep here necessary? Should we build in 1 or 2 rechecks if the abnormal calls fail?

PR #501, opened 2023-08-15
Add dynamic baselining to Jenkins job

nathan-weinberg:
Resolves [NETOBSERV-1239](https://issues.redhat.com/browse/NETOBSERV-1239)


nathan-weinberg:
Initially tested here: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/nweinber-e2e-benchmarking-multibranch-pipeline/job/netobserv-perf-tests/379/

The fetching succeeded, but the upload remains untested as the Tolerancy Check failed (although the control flow did perform as expected)

Perhaps we should update our Tolerancy Config as part of this PR, as this will be the last update before cron scheduling, which will be configured via GitLab. Thoughts on that?

scripts/netobserv.sh
```diff
@@ -110,7 +110,7 @@ deploy_lokistack() {
   elif [[ $LOKISTACK_SIZE == "1x.medium" ]]; then
     LokiStack_CONFIG=$SCRIPTS_DIR/loki/lokistack-1x-medium.yaml
   else
-    echo "====> No LokiStack config was found - using 1x.extra-small"
+    echo "====> No LokiStack config was found - using '1x.extra-small'"
     echo "====> To set config, set LOKISTACK_SIZE variable to either '1x.extra-small', '1x.small', or '1x.medium'"
```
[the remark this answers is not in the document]
nathan-weinberg: Yup! Going to need to introduce the logic to determine whether or not we want `extra-small` or `demo` depending on Loki Operator version, but out-of-scope for this PR.

scripts/queries/netobserv_touchstone_tolerancy_rules.yaml
```diff
@@ -1,21 +1,21 @@
...
+  tolerancy: 10
   max_failures: 0
 - json_path: ["metric_name", "rssEBPFTotals", "metric_name", "*", "avg(value)"]
-  tolerancy: 5
+  tolerancy: 10
   max_failures: 0
 - json_path: ["metric_name", "rssFLPTotals", "metric_name", "*", "avg(value)"]
-  tolerancy: 5
+  tolerancy: 10
   max_failures: 0
```
memodi, line 21: could we add Loki and Kafka CPU, RSS totals as well? it would be very high tolerance value, say 60%, wdyt?

scripts/analyze_metrics.py
```diff
@@ -1,73 +0,0 @@
-#!/usr/bin/env python3
-
-import sys
-import json
```
memodi, line 4: ah, I had forgotten about this script. good old times when we used to analyze data with this :D Long Live touchstone!!

Jenkinsfile
```diff
@@ -290,13 +290,21 @@ pipeline {
...
         string(
-            name: 'BASELINE_UUID',
+            name: 'BASELINE_UUID_OVERRIDE',
             defaultValue: '',
             description: '''
-                Baseline UUID to compare this run to<br/>
-                If the workload completes successfully and is uploaded to Elasticsearch, adding a UUID here will configure Touchstone to run a comparison between the job run and the given baseline UUID<br/>
-                If no UUID is specified, no comparison will be made
+                Override baseline UUID to compare this run to<br/>
+                Note you will still need to check the box above to run the comparison
```
memodi, line 307: why RUN_BASELINE_COMPARISON checkbox needs to be checked, could we simply do comparison if UUID is specified?

memodi (changes requested):
looks good overall. I think we should hold off though as you suggested to wait for final approval for tolerancy configs from devs next week.

scripts/queries/netobserv_touchstone_tolerancy_rules.yaml
nathan-weinberg: Yes can do!

Jenkinsfile
nathan-weinberg: Checkbox determines if we want to run a comparison or not - if a UUID override is not provided the job uses the NOPE tool to dynamically fetch one for whatever workload was run

scripts/queries/netobserv_touchstone_tolerancy_rules.yaml
```diff
@@ -1,21 +1,36 @@
+# KPIs
 - json_path: ["metric_name", "nFlowsProcessedPerMinuteTotals", "metric_name", "*", "avg(value)"]
-  tolerancy: 10
+  tolerancy: 0
   max_failures: 0
 - json_path: ["metric_name", "nFlowsErroredPerMinuteTotals", "metric_name", "*", "avg(value)"]
   tolerancy: 0
   max_failures: 0
 - json_path: ["metric_name", "lokiRecordsDropped", "metric_name", "*", "avg(value)"]
-  tolerancy: 2
```
memodi, line 8: thanks for updating these values, do you want to create a task to update nope tool to update with 0 value if nothing returns from Prometheus?

nathan-weinberg: @memodi let's do as a followup, this PR is already quite large, I'll create a new task under the Enhancements story

memodi: in the mean time we should continue to pay attention regular spreadsheet for flow errors/drops metrics.

PR #502, opened 2023-08-21
Adding options for moving pods to infra nodes in cluster post config

liqcui:
Adding options for moving pods to infra nodes in cluster post config and support multi-arch

The Jenkins Logs:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/liqcui-e2e-benchmarking-multibranch-pipeline/job/cluster-post-config/223/console

PR #503, opened 2023-08-21
add checkbox and default not moving ingress monitoring and registry p…

qiliRedHat:
…ods to infra nodes
Depends on https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/502 to merge


paigerube14:
@qiliRedHat what is the reasoning here of not wanting to move monitoring for ingress tests? 
Before this merges we will also need an update into cluster-worker-scaling to add the parameters as well

qiliRedHat:
> @qiliRedHat what is the reasoning here of not wanting to move monitoring for ingress tests? Before this merges we will also need an update into cluster-worker-scaling to add the parameters as well

@paigerube14 
We want only ingress pod on infra nodes. So monitoring is not moved. Ingress-perf itself will [move](https://github.com/cloud-bulldozer/e2e-benchmarking/blob/master/workloads/ingress-perf/config/standard.yml#L12) the ingress pod to infra nodes, so we don't have to move ingress here too.
As in the cluster-worker-scaling and cluster-post-config jobs, all the 3 parameters are default to true, so it should not impact any other workloads, will move monitoring/ingresss/registry by default if not explicitly passing them as false like in ingress-perf.

You're right, update into cluster-worker-scaling https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/507 is merged

qiliRedHat:
@paigerube14 PTAL

PR #517, opened 2023-09-07
Added additional exception logging to Mr. Sandman

nathan-weinberg:
Also did some whitespace cleanup


nathan-weinberg:
At @paigerube14's request I've also added colored logging to Mr. Sandman (and removed the `time` library which seems to be unused)

PR #520, opened 2023-09-12
Cluster post config

paigerube14:
Separating out moving components. 
This helps if the infra nodes were created but you know want to be able to move the ingress, monitoring or registry 

Infra pods had been created in previous step but wanted to be able to move the components but wasn't able to 

https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/cluster-post-config/1629/console

Passing job; https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/cluster-post-config/118/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -145,10 +145,6 @@ pipeline {
 
               SECONDS=0
               ./openshift-qe-workers-infra-workload-commands.sh
-              export $IF_MOVE_INGRESS
-              export $IF_MOVE_REGISTRY
-              export $IF_MOVE_MONITORING
-              ./openshift-qe-move-pods-infra-commands.sh
               status=$?
               echo "final status $status"
               duration=$SECONDS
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -166,6 +166,60 @@ pipeline {
...
+              set -a && source .env_override && set +a
+              mkdir -p ~/.kube
+              cp $WORKSPACE/flexy-artifacts/workdir/install-dir/auth/kubeconfig ~/.kube/config
+
+              oc config view
+              oc projects
+              ls -ls ~/.kube/
+              env
+              set -x
+              ./openshift-qe-move-pods-infra-commands.sh
```
qiliRedHat, line 205: @paigerube14 I think this script already have run in line 151 if (params.IF_MOVE_INGRESS == true ) || (params.IF_MOVE_REGISTRY == true) || (params.IF_MOVE_MONITORING == true ) in the previous stage.
To avoid the script to be run twice and cause issue, the new stage you added should have one more and condition (params.INFRA_NODES == false). Or we can remove lines 147-151.

paigerube14: thanks! forgot to remove that section earlier

PR #545, opened 2023-10-26
update code to extend the timeout of checking upgrade

liqcui:
update code to extend the timeout of checking upgrade and ignore terminating pod
Jenkins Logs:
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/liqcui-e2e-benchmarking-multibranch-pipeline/job/loaded-upgrade/110/console

Adding UPGRADE_WAIT_NUM to extend the timeout of checking check_upgrade.check_upgrade 

because ibmcloud will take more time to complete loaded upgrade, so we need to increase the timeout of check_upgrade.check_upgrade, the default value 300 isn't enough

### DIFF
--- a/upgrade_scripts/common.sh
+++ b/upgrade_scripts/common.sh
@@ -146,7 +146,7 @@ function capture_failed_pods_before_upgrade(){
   #Save the failed job before upgrade and make sure new failed job caused by upgrade,ignore installer, build error
   echo "Capture failed pods before upgrade OCP"
   echo "####################################################################################"
-  oc get pods -A| grep -v -E 'Running|Completed|installer|build'| awk '(NF=NF-2) 1'>/tmp/upgrade-before-failed-pods.txt
+  oc get pods -A| grep -v -E 'Running|Completed|installer|build|Terminating'| awk '(NF=NF-2) 1'>/tmp/upgrade-before-failed-pods.txt
 }
 
 function capture_failed_pods_after_upgrade(){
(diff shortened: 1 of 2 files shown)


liqcui:
@qiliRedHat @paigerube14 please review my PR when you are available, thanks!

upgrade_scripts/common.sh
```diff
@@ -146,15 +146,15 @@ function capture_failed_pods_before_upgrade(){
   #Save the failed job before upgrade and make sure new failed job caused by upgrade,ignore installer, build error
   echo "Capture failed pods before upgrade OCP"
   echo "####################################################################################"
-  oc get pods -A| grep -v -E 'Running|Completed|installer|build'| awk '(NF=NF-2) 1'>/tmp/upgrade-before-failed-pods.txt
+  oc get pods -A| grep -v -E 'Running|Completed|installer|build|Terminating'| awk '(NF=NF-2) 1'>/tmp/upgrade-before-failed-pods.txt
```
paigerube14, line 149: I think we will want to fail if pods are stuck in terminating here

liqcui: Actually this is for pods in openshift-marketplace, the pod will be re-create periodically, it's not bug, if we want to check if it's really stuck in terminating state, we need to add more sleep time to wait.

liqcui: Removing Terminating and adding sleep before capture pod status, thanks!

qiliRedHat:
@liqcui lgtm. Leave to @paigerube14 to review your change for her comment.

liqcui:
@paigerube14 anymore comments?

PR #552, opened 2023-11-16
Pre-merge image patching

Amoghrd:

PR #553, opened 2023-11-21
Fixed bug with image patching

nathan-weinberg:
Fixing bug introduced in https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/552

Job fails if no patching is done due to how variables are handled, see https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/nweinber-e2e-benchmarking-multibranch-pipeline/job/netobserv-perf-tests/448/console

This moves the patch check into the `if` block itself

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,13 +10,17 @@ if (userId) {
 }
 
 pipeline {
+
+    // job runs on specified agent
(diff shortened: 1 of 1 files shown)


Amoghrd:
Thanks for the fix
/lgtm

PR #576, opened 2024-01-23
Add additional K8S object tracking to Touchstone configs

nathan-weinberg:
Resolves [NETOBSERV-1252](https://issues.redhat.com/browse/NETOBSERV-1252)


memodi (commented):
I think for these metrics instead of avg just 'max' value is what we need? percentile is also not required.

memodi:
thanks @nathan-weinberg.
/lgtm

PR #582, opened 2024-02-08
Add ability to change e2e-benchmarking repo and branch

nathan-weinberg:
For additional flexibility, tested here: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/nweinber-e2e-benchmarking-multibranch-pipeline/job/netobserv-perf-tests/538/

### DIFF
--- a/scripts/netobserv/flows_v1beta2_flowcollector.yaml
+++ b/scripts/netobserv/flows_v1beta2_flowcollector.yaml
@@ -44,7 +44,7 @@ spec:
       minReplicas: 1
       status: Disabled
     imagePullPolicy: IfNotPresent
-    logLevel: debug
+    logLevel: info
     portNaming:
       enable: true
       portNames:
--- a/scripts/netobserv/flows_v1beta2_flowcollector.yaml
+++ b/scripts/netobserv/flows_v1beta2_flowcollector.yaml
@@ -11,7 +11,6 @@ spec:
       excludeInterfaces:
       - lo
       imagePullPolicy: IfNotPresent
-      kafkaBatchSize: 10485760
       logLevel: info
       resources:
         limits:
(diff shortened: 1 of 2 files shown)


nathan-weinberg:
cc @jotak @msherif1234 once this PR is merged all QE perf tests will use the lower `kafkaBatchSize` regardless of operator default (since we are overriding via FC)

jotak:
@nathan-weinberg sounds good ... curious to see a ingress-perf run, it's the only one that we haven't checked for this afaik?

memodi:
@nathan-weinberg - how about removing this option altogether so we test with whatever the default Operator configures? this is also in sync with our other theme to use defaults wherever possible unless necessary to use custom value

scripts/netobserv/flows_v1beta2_flowcollector.yaml
```diff
@@ -11,7 +11,7 @@ spec:
       excludeInterfaces:
       - lo
       imagePullPolicy: IfNotPresent
-      kafkaBatchSize: 10485760
+      kafkaBatchSize: 1048576
```
memodi, line 14: @nathan-weinberg - how about removing this option altogether so we test with whatever the default Operator configures? this is also in sync with our other theme to use defaults wherever possible unless necessary to use custom value

nathan-weinberg:
> @nathan-weinberg sounds good ... curious to see a ingress-perf run, it's the only one that we haven't checked for this afaik?

I'll be doing runs for all the workloads with the 1.5 GA image

scripts/netobserv/flows_v1beta2_flowcollector.yaml
nathan-weinberg: yeah, i'm open to that

PR #584, opened 2024-02-26
Force NetObserv DOWNSTREAM_DEPLOYMENT to always be true

nathan-weinberg:
As discussed [here](https://redhat-internal.slack.com/archives/C03SHCD5C9M/p1708973082567179) we have recently been having issues with authenticating Dittybopper and the NOPE tool with the Upstream operator - however, since we now have the `DOWNSTREAM_DEPLOYMENT` flag in the operator CSV, a potential mitigation for this problem is simply always setting that flag to `true` since we only ever test on OpenShift and not vanilla Kubernetes. This has the added benefit of simplifying a lot of our codebase as we no longer need to differentiate between the upstream and downstream operator for metric collection. 

Would resolve [NETOBSERV-1524](https://issues.redhat.com/browse/NETOBSERV-1524)


memodi (approved):
/lgtm

PR #588, opened 2024-03-20
Periodic job - gsheet generation

paigerube14:
Storing periodic job generation document in prow folder

### DIFF
--- a/prow/get_periodic_jobs.py
+++ b/prow/get_periodic_jobs.py
@@ -16,13 +16,13 @@ def invoke(command):
         print("Failed to run %s" % (command))
     return output
 
-def add_new_worksheet(row): 
+def add_new_worksheet(row,sheet_loc): 
     scope = [
         "https://spreadsheets.google.com/feeds",
         "https://www.googleapis.com/auth/drive",
     ]
     credentials = ServiceAccountCredentials.from_json_keyfile_name(
-        "/Users/prubenda/.secrets/perf_sheet_service_account.json", scope
+       sheet_loc, scope
     )
     
     gc = gspread.authorize(credentials)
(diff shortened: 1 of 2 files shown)


prow/get_periodic_jobs.py
```diff
@@ -0,0 +1,233 @@
...
+    except Exception:
+        print("Failed to run %s" % (command))
+    return output
+
+def add_new_worksheet(row): 
+    scope = [
+        "https://spreadsheets.google.com/feeds",
+        "https://www.googleapis.com/auth/drive",
+    ]
+    credentials = ServiceAccountCredentials.from_json_keyfile_name(
```
pruan-rht, line 24: @paigerube14 it's locked to your localhost's path?

prow/get_periodic_jobs.py
```diff
@@ -0,0 +1,233 @@
...
+    return final_row
+
+final_csv = "periodic.csv"
+# invoke("rm -rf release_master")
+# invoke("mkdir release_master")
+# invoke("cd release_master; git clone https://github.com/openshift/release.git")
+#folder_path="./release_master/release/ci-operator/config/openshift-qe/ocp-qe-perfscale-ci"
+
+
+folder_path="/Users/prubenda/PycharmProjects/release/ci-operator/config/openshift-qe/ocp-qe-perfscale-ci"
```
pruan-rht, line 219: @paigerube14 this would mean it has to be run from your laptop...perhaps think about changing that.

prow/get_periodic_jobs.py
paigerube14: yes, this file is only used locally for someone to use and run. We dont have a need to run on jenkins. We just needed a better place to store this file for people to have access to and run

pruan-rht: @paigerube14 perhaps use an environment variable so that it's configurable.  Another option is to use store it in bitwarden and then use the bitwarden API to access the creds.

svetsa-rh:
/lgtm

Chatted with Paige. We decided that it might be best to upload current one now to folks can start using it and make changes suggested by @pruan-rht or other changes at a later time.

PR #593, opened 2024-04-04
always dump, upload when dump is missing

willr3:
NOPE currently dumps to file or uploads to elastic search. I propose we always dump to file and use the `--dump` to disable elastic search uploading if desired. This will allow us to send data to elastic search and Horreum from the same performance test instead of having to choose which data processing will be performed when scheduling the test


willr3:
This PR changes NOPE to always dump to file (currently only active when using `--dump`). The `--dump` option would still disable the elasticsearch upload to maintain that part of its current functionality.

nathan-weinberg:
@willr3 I think we should rename the flag - I don't think it makes sense in it's current form - maybe `--dump-only`?

This is also going to require updated to the command line arg help text, the `Jenkinsfile`, and the `README`

willr3:
Sure. I will update the PR to include a rename to the flag. I kept the original name to avoid changing other parts of the workflow. The current use of `--dump` already equated to disabling the elastic-search upload but `--dump-only` is an easier to understand option for users.

willr3:
I updated the PR to use `--dump-only` based on what I saw when looking at `args` and other parameters that use `-`. I also found another place outside of `main()` where the `DUMP` environment variable was being read so hopefully I updated that correctly.

nathan-weinberg:
@willr3 what you have looks good but still need updates to the pieces I mentioned above

> This is also going to require updates to the command line arg help text, the Jenkinsfile, and the README

willr3:
It took me a bit longer than I would like to realize that the other files you were mentioning were also in this repository :)
I updated the Jenkinsfile and `scripts/README.md` wherever I saw `--dump` or the dump option mentioned. Please let me know if I missed anything

Jenkinsfile
```diff
@@ -329,7 +329,7 @@ pipeline {
             name: 'NOPE_DUMP',
             defaultValue: false,
             description: '''
-                Check this box to dump data collected by the NOPE tool to a file instead of uploading to Elasticsearch<br/>
+                Check this box to dump data collected by the NOPE tool to a file without uploading to Elasticsearch<br/>
```
nathan-weinberg, line 332: Can you change the `name:` of this Jenkins field to be `NOPE_DUMP_ONLY`

scripts/nope.py
```diff
@@ -530,10 +530,10 @@ def main():
...
     # either dump data locally or upload it to Elasticsearch
     timestamp = datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')
-    if DUMP:
-        dump_data_locally(timestamp)
-        logging.info(f"Data written to {DATA_DIR}/data_{timestamp}.json")
-    else:
+    
+    dump_data_locally(timestamp)
+    logging.info(f"Data written to {DATA_DIR}/data_{timestamp}.json")
+    if not DUMP:
```
nathan-weinberg, line 536: Can you change the name of var to `DUMP_ONLY`

nathan-weinberg (approved):
LGTM just two nits and we can merge!

Jenkinsfile
willr3: Absolutely, I will update and push shortly

PR #597, opened 2024-04-29
Enable metrics under eBPF FC config

nathan-weinberg:


memodi:
@nathan-weinberg - processor default metrics port also changed to 9401, can you update that as well?

nathan-weinberg:
@memodi https://github.com/openshift-qe/ocp-qe-perfscale-ci/pull/598

PR #605, opened 2024-06-07
Fix metric promql: Bps measured by netobserv

jotak:
Changing this promQL for 2 things:
- it was incorrectly looking at traffic going **from** the workloads, but the analog cAdvisor query looks at traffic going **to** the workloads (`container_network_receive_bytes_total{job="kubelet", metrics_path="/metrics/cadvisor", cluster="", namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*"}`). So we need to change `SrcK8S_Namespace` to `DstK8S_Namespace`.
- changing `DstK8S_OwnerType="Deployment"` to `DstK8S_Type!="Service"`. The purpose is the same: eliminate duplicates resulting from the services layer. The new filters are more accurate.

### DIFF
--- a/scripts/queries/netobserv_prometheus_queries.yaml
+++ b/scripts/queries/netobserv_prometheus_queries.yaml
@@ -19,7 +19,7 @@
   metricName: nWorkloadFlowsProcessedPerMinuteTotals
 
 # total bytes processed from workload namespaces by netobserv
-- query: sum(rate(netobserv_workload_ingress_bytes_total{SrcK8S_Namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*", DstK8S_OwnerType="Deployment"}[1m])*60)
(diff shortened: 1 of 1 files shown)


jotak:
actually, hold on, I'm doing more tests to try to better align our queries with cadvisor queries
/hold

jotak:
Ok this should be aligned now, hopefully. E.g. same queries ran against some random openshift namespace:

image

memodi:
/ok-to-test

LGTM, I'll let @nathan-weinberg add final approval to merge since he has been running perf tests for 1.6.

nathan-weinberg:
@memodi @Amoghrd final labels will be needed from y'all as I am no longer an [OWNER](https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/netobserv-perf-tests/OWNERS)

PR #606, opened 2024-06-13
NETOBSERV-1694 metrics and fixing kafkaTopic delete issue

memodi:
ebpf metrics, enabling FLP metrics and fixing kafkaTopic delete issue

### DIFF
--- a/scripts/netobserv/flows_v1beta2_flowcollector.yaml
+++ b/scripts/netobserv/flows_v1beta2_flowcollector.yaml
@@ -39,6 +39,10 @@ objects:
         multiClusterDeployment: false
         addZone: false
         metrics:
+          includeList:
+            - node_ingress_bytes_total
+            - workload_ingress_bytes_total
+            - namespace_flows_total
           server:
             port: 9401
             tls:
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -136,6 +136,11 @@ The NOPE tool can also be used for fetching and uploading baselines on a workloa
 ## Fetching metrics using Touchstone 
 NetObserv metrics uploaded to Elasticsearch can be fetched using `touchstone` tool provided by [benchmark-comparison](https://github.com/cloud-bulldozer/benchmark-comparison). Once you have Touchstone set up, you can run it with any given UUID using the `netobserv_touchstone_statistics_config.json` file in the `queries/` directory under `scripts/`
 
+## Different touchstone configs in this branch:
+- [netobserv_touchstone_statistics_config.json](./queries/netobserv_touchstone_statistics_config.json) - collect *all* metrics including per pod, typically used for producing raw datasheets.
+- [netobserv_touchstone_tolerancy_config.json](./queries/netobserv_touchstone_tolerancy_config.json) - collects only *totals* of each metric, typically used for comparison datasheet.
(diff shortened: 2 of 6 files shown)


scripts/netobserv/flows_v1beta2_flowcollector.yaml
```diff
@@ -39,6 +39,10 @@ objects:
         multiClusterDeployment: false
         addZone: false
         metrics:
+          includeList:
```
Amoghrd, line 42: Think if you dont add includeList param in flowcollector. There are default values. Think these are those default values. Is it needed to explicitly add these?

memodi: yes, we need this for some metrics we are collecting. I missed this in my previous PR where I converted to template and some metrics were not populated in spreadsheet.

Amoghrd: so these metrics are not default?

memodi: no, workload_flows_total is default

smg247:
https://issues.redhat.com/browse/NETOBSERV-1694

scripts/netobserv/flows_v1beta2_flowcollector.yaml
jotak: Sorry I'm late here
I think we could still use the default values and manage the different Loki cases. 

- When Loki is enabled, by default `netobserv_namespace_flows_total` is used
- When Loki is disabled, by default `netobserv_workload_flows_total` is used

To make both cases work, we can adapt the promql changing this:

```yaml

# total flows processed from workload namespaces
- query: sum(netobserv_namespace_flows_total{SrcK8S_Namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*"})
  metricName: nWorkloadFlowsProcessedTotals

# total flows processed from workload namespaces per minute
- query: sum(rate(netobserv_namespace_flows_total{SrcK8S_Namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*"}[1m])*60)
  metricName: nWorkloadFlowsProcessedPerMinuteTotals

```

for that:

```yaml

# total flows processed from workload namespaces
- query:
    sum(netobserv_namespace_flows_total{SrcK8S_Namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*"})
    or sum(netobserv_workload_flows_total{SrcK8S_Namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*"})
  metricName: nWorkloadFlowsProcessedTotals

# total flows processed from workload namespaces per minute
- query:
    sum(rate(netobserv_namespace_flows_total{SrcK8S_Namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*"}[1m])*60)
    or sum(rate(netobserv_workload_flows_total{SrcK8S_Namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*"}[1m])*60)
  metricName: nWorkloadFlowsProcessedPerMinuteTotals

```

jotak: (the advantage is to make sure we always test the default values, including when we add more default metrics, like it's the case in lokiless, as it can impact FLP resource usage)

memodi: sure, thanks, I'll add them in for 1.7 runs. For now, I'd like to keep the environment and metrics same as what we had for runs we had with Loki enabled.

PR #620, opened 2024-07-15
FLP kafka replicas bug fix

memodi:
FLP kafka replicas bug fix


Amoghrd:
Oh my bad with the if condition
/lgtm

PR #645, opened 2024-09-16
NETOBSERV-1878 use loki 6.0 channel for OCP >= 4.17

memodi:
[NETOBSERV-1878](https://issues.redhat.com//browse/NETOBSERV-1878) use loki 6.0 channel for OCP >= 4.17

### DIFF
--- a/scripts/loki/loki-released-subscription.yaml
+++ b/scripts/loki/loki-released-subscription.yaml
@@ -4,7 +4,7 @@ metadata:
   name: loki-operator
   namespace: openshift-operators-redhat
 spec:
-  channel: stable
+  channel: stable-6.0
   name: loki-operator
   source: redhat-operators
   sourceNamespace: openshift-marketplace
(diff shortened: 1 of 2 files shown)


scripts/loki/loki-released-subscription.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: loki-operator
   namespace: openshift-operators-redhat
 spec:
-  channel: stable
+  channel: stable-6.0
```
Amoghrd, line 7: since loki-6.0 is not yet released, it should only be for unreleased subscription right

memodi: since it's going to be GA soon (Sept 24) I included for released as well. And for cron jobs, we're going to us `Unreleased` option: https://gitlab.cee.redhat.com/aosqe/jenkins-jcasc-n/-/merge_requests/2340

Amoghrd: okay makes sense. Hope they keep the same channel name post release too and we dont have to change it back to `stable`

PR #654, opened 2024-10-09
Changes for installing loki < 6.0 with the stable channel on OCP < 4.17

AdityaHonkalas:
Added the changes to install Loki operator based on the OCP version.


scripts/loki/loki-released-subscription.yaml
memodi, line 1: having 2 identical template is not needed, we could simply delete one of them and name "loki-subscription-template.yaml"

scripts/loki/loki-released-subscription.yaml
```diff
@@ -1,10 +1,18 @@
...
+objects:
+- apiVersion: operators.coreos.com/v1alpha1
+  kind: Subscription
+  metadata:
+    name: loki-operator
+    namespace: openshift-operators-redhat
+  spec:
+    channel: "${{LokiChannel}}"
+    name: loki-operator
+    source: redhat-operators
```
memodi, line 14: could add a variable for source as well

scripts/netobserv.sh
```diff
@@ -105,12 +105,24 @@ deploy_lokistack() {
...
+  then
+    loki_channel="stable-6.0" ; echo "=====> Deploying the Loki with stable-6.0 channel..."
+  else
+    loki_channel="stable" ; echo "=====> Deploying the Loki with the stable channel..."
+  fi
+
   echo "====> Creating netobserv-downstream-testing CatalogSource (if applicable) and Loki Operator Subscription"
   if [[ $LOKI_OPERATOR == "Released" ]]; then
-    oc apply -f $SCRIPTS_DIR/loki/loki-released-subscription.yaml
+    oc process --ignore-unknown-parameters=true -f $SCRIPTS_DIR/loki/loki-released-subscription.yaml -p LokiChannel=$loki_channel -o yaml > /tmp/loki-released-subscription.yaml
```
memodi, line 120: use the same template file and set source to be "redhat-operators"

scripts/netobserv.sh
```diff
@@ -105,12 +105,24 @@ deploy_lokistack() {
...
+    oc apply -f /tmp/loki-released-subscription.yaml -n openshift-operators-redhat
   elif [[ $LOKI_OPERATOR == "Unreleased" ]]; then
     deploy_downstream_catalogsource
-    oc apply -f $SCRIPTS_DIR/loki/loki-unreleased-subscription.yaml
+    oc process --ignore-unknown-parameters=true -f $SCRIPTS_DIR/loki/loki-unreleased-subscription.yaml -p LokiChannel=$loki_channel -o yaml > /tmp/loki-unreleased-subscription.yaml
```
memodi, line 124: use the same template file and set source to be "netobserv-downstream-testing"

memodi (changes requested):
thanks @AdityaHonkalas for the PR.

scripts/loki/loki-released-subscription.yaml
AdityaHonkalas: Added a single subscription template for both released/ unreleased version installation.

scripts/loki/loki-released-subscription.yaml
AdityaHonkalas: Added SOURCE var to the template.

scripts/netobserv.sh
AdityaHonkalas: Done.

PR #655, opened 2024-10-14
While loop to wait for loki-operator and bug fix in Jenkinsfile

memodi:
While loop to wait for loki-operator and bug fix in Jenkinsfile

this should prevent errors like:
```
10-14 07:31:12.867  + oc wait --timeout=180s --for=condition=ready pod -l app.kubernetes.io/name=loki-operator -n openshift-operators-redhat
10-14 07:31:12.867  error: no matching resources found
```
and 
```
10-14 11:46:42.726  /home/jenkins/ws/workspace/ch-pipeline_netobserv-perf-tests@tmp/durable-202ebc20/script.sh.copy: line 4: syntax error near unexpected token `)'
```
latter causes comparison to fail and marking whole run as Unstable.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -888,7 +888,6 @@ pipeline {
                                     baselineReturnCode = sh(returnStatus: true, script: """
                                             source $WORKSPACE/ocp-qe-perfscale-ci/scripts/run_py_scripts.sh
                                             run_comparison
-                                        )
                                     """)
 
                                     if(fileExists("${BENCHMARK_COMP_LOG}")){
--- a/scripts/netobserv.sh
+++ b/scripts/netobserv.sh
@@ -135,7 +135,10 @@ deploy_lokistack() {
   echo "====> Creating S3 secret for Loki"
   $SCRIPTS_DIR/deploy-loki-aws-secret.sh $S3_BUCKET_NAME
   sleep 60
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -888,7 +888,6 @@ pipeline {
                                     baselineReturnCode = sh(returnStatus: true, script: """
                                             source $WORKSPACE/ocp-qe-perfscale-ci/scripts/run_py_scripts.sh
                                             run_comparison
-                                        )
```
paigerube14, line 891: did you mean to take out this parentheses?

Amoghrd:
I dont see the reason for a while loop still. The wait itself is running a for loop waiting for the pod to come up
Shouldnt just increasing the timeout be enough?

memodi:
> I dont see the reason for a while loop still. The wait itself is running a for loop waiting for the pod to come up Shouldnt just increasing the timeout be enough?

oc wait only helps if the pod is scheduled but not yet running, while loop will help when the pod is not yet scheduled and pod is not yet created. We're already waiting for 60 seconds, increasing timeout is more of a guess work in a hope it will succeed but it's not deterministic.

Amoghrd:
okay makes sense. I thought increasing the sleep time before this command would also help but as you said before its not deterministic
/lgtm

PR #659, opened 2024-10-29
NETOBSERV-1874 Netobserv reliability tests

memodi:
[NETOBSERV-1874](https://issues.redhat.com//browse/NETOBSERV-1874) Netobserv reliability tests
the code changes here are improvements in general, should help reliability tests and also other performance tests.
- source env.sh should have all necessary env. vars for full stack installation. 
- defaults for some of variables.
- [NETOBSERV-1961](https://issues.redhat.com//browse/NETOBSERV-1961) have ebpf privileged mode by default.
- adds dashboards for file descriptor and goroutine thread to check for leaks.

### DIFF
--- a/scripts/env.sh
+++ b/scripts/env.sh
@@ -1,8 +1,6 @@
-# Should be 'Official', 'Internal', 'OperatorHub' or 'Source'
-export INSTALLATION_SOURCE=Internal
 # will use 'Released' if not set otherwise
 export LOKI_OPERATOR=Unreleased
 # will use '1x.extra-small' if not set 
-export LOKISTACK_SIZE=1x.small
+export LOKISTACK_SIZE=1x.extra-small
 # will use '6' if not set 
 export TOPIC_PARTITIONS=48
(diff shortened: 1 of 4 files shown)


memodi:
@qiliRedHat - what would be good place to add a cleanup function for netobserv? I would like to simply call [this function](https://github.com/openshift-qe/ocp-qe-perfscale-ci/blob/netobserv-perf-tests/scripts/netobserv.sh#L299) when reliability test env is also cleared up.

memodi:
added in e297dec, PTAL.

memodi:
@jotak @Amoghrd this is ready for review

scripts/env.sh
```diff
@@ -0,0 +1,6 @@
+# will use 'Released' if not set otherwise
+export LOKI_OPERATOR=Unreleased
```
Amoghrd, line 2: Why are we using Unreleased loki? For our perf tests we use Released Loki as far as I remember. 
Using Unreleased might cause its own issues right?

memodi: 6.0 released loki is not supported with 4.18 OCP version, so had to use unreleased 6.1

Amoghrd: oh okay gotcha

Amoghrd:
LGTM

memodi:
thanks @Amoghrd - I think this is good to get merged.

PR #660, opened 2024-10-31
Cluster post config

paigerube14:


paigerube14:
Successful job: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/paige-e2e-multibranch/job/cluster-post-config/128/console

/assign @memodi

openshift-qe-move-pods-infra-commands.sh
```diff
@@ -483,7 +483,7 @@ echo "Final check - Check if all pods to be settle"
 sleep 5
 max_retries=30
 retry_times=1
-while [[ $(oc get pods --no-headers -n openshift-monitoring | grep -Pv "(Completed|Running)" | wc -l) != "0" ]];
+while [[ $(oc get pods --no-headers -n openshift-monitoring | grep -e "(Completed|Running)" | wc -l | xargs) != "0" ]];
```
memodi, line 486: are you looking for pods that are in Completed Or Running state? or you want to exclude Pods that are in Completed Or Running state ? Looks like in previous code you were trying to do latter.

paigerube14: Based on the echo within the if and while statement I think we want to find if there are any pods not in running or completed state. 

I'm confused as to how it worked before now

memodi: I think you want to use `-E` option of grep and -v (like you had before), you also don't need parentheses for multiple regex pattern with `-E`

```
$(oc get pods --no-headers -n openshift-monitoring | grep -Ev "Completed|Running" | wc -l | xargs)
```

memodi (approved):
/lgtm

PR #663, opened 2024-11-08
NETOBSERV-1962 Run ingress-perf in parallel with NDH and CD workloads

memodi:
- [NETOBSERV-1962](https://issues.redhat.com/browse/NETOBSERV-1962) Run ingress-perf in parallel with NDH and CD workloads
- [NETOBSERV-1963](https://issues.redhat.com/browse/NETOBSERV-1963) Enable kube-burner-ocp Churn option for node-density-heavy and cluster-density-v2 workloads -- currently Churn option is only enabled for cluster-density workloads. For NDH, it's dependent on the PR: https://github.com/kube-burner/kube-burner-ocp/pull/93
- Option to run ingress-perf in parallel (Default: true).  When unchecked we would be able to run NDH and CD-v2 as standalone.
- Option to run ingress-perf workload as standalone.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,6 +16,8 @@ pipeline {
         def BENCHMARK_CSV_LOG = "$WORKSPACE/e2e-benchmarking/benchmark_csv.log"
         def BENCHMARK_COMP_LOG = "$WORKSPACE/e2e-benchmarking/benchmark_comp.log"
         def templateParams = ''
+        def ingressWorkloadJob = ''
+        def kubeburnerWorkloadJob = ''
     }
 
     // job runs on specified agent
@@ -25,6 +27,7 @@ pipeline {
     options {
         timeout(time: 6, unit: 'HOURS')
         ansiColor('xterm')
+        parallelsAlwaysFailFast()
     }
 
     // job parameters
@@ -62,8 +65,8 @@ pipeline {
         )
         booleanParam(
             name: 'INFRA_WORKLOAD_INSTALL',
-            defaultValue: false,
-            description: 'Install workload and infrastructure nodes even if less than 50 nodes'
+            defaultValue: true,
+            description: 'Install workload and infrastructure nodes; required for ingress-perf. Uncheck if you already have infra nodes in your cluster'
         )
         booleanParam(
             name: 'INSTALL_DITTYBOPPER',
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -689,74 +692,127 @@ pipeline {
...
                 )
                 script {
                     // set new env vars from workload 'index_data' JSON file and update build description fields
-                    workloadInfo = readJSON(file: "$WORKSPACE/workload-artifacts/index_data.json")
-                    workloadInfo.each { env.setProperty(it.key.toUpperCase(), it.value) }
-                    // UUID
+                    kubeburnerWorkloadInfo = readJSON(file: "$WORKSPACE/kube-burner-workload-artifacts/index_data.json")
+                    ingressperfWorkloadInfo = readJSON(file: "$WORKSPACE/ingressperf-workload-artifacts/index_data.json")
+                    kubeburnerStarttime = kubeburnerWorkloadInfo.startDateUnixTimestamp
+                    ingressperfStarttime = ingressperfWorkloadInfo.startDateUnixTimestamp
```
jotak, line 793: what do you think about having a delay of a few minutes before starting ingress-perf?
I think it could help if we need to troubleshoot something (did an issue appear before or after ingress-perf started)

Amoghrd:
code LGTM

Jenkinsfile
memodi: thanks, I added sleep of 10 mins before ingress-perf workload is triggered in parallel.

Jenkinsfile
```diff
@@ -1055,9 +1071,6 @@ def setKafkaReplicas(){
         if (params.WORKLOAD == 'node-density-heavy') {
             env.FLP_KAFKA_REPLICAS = '6'
         }
-        else if (params.WORKLOAD == 'ingress-perf') {
```
Amoghrd, line 1058: This can be kept I think. If ingress-perf is run standalone then Kafka_Replicas need to be set to 12

Jenkinsfile
```diff
@@ -688,79 +697,87 @@ pipeline {
...
-                            string(name: 'IMAGE', value: NETOBSERV_MUST_GATHER_IMAGE),
-                            string(name: 'JENKINS_AGENT_LABEL', value: params.JENKINS_AGENT_LABEL),
-                            booleanParam(name: 'GEN_CSV', value: false),
-                            string(name: 'E2E_BENCHMARKING_REPO', value: params.E2E_BENCHMARKING_REPO),
-                            string(name: 'E2E_BENCHMARKING_REPO_BRANCH', value: params.E2E_BENCHMARKING_REPO_BRANCH)
-                        ]
+            parallel {
+                stage('kube-burner workload'){
+                    when {
+                        expression { params.WORKLOAD != 'ingress-perf' }
```
Amoghrd, line 707: wont this if condition be satisfied even if workload = None? Think better to add if condition to check for node-desnity and cluster-density particularly

Jenkinsfile
```diff
@@ -1087,3 +1100,90 @@ def setTemplateParams(){
     }
     return templateParams
 }
+
+def captureWorkloadResults(){
+    def UUID = ''
+    def JENKINS_BUILD = ''
+
+    if (params.WORKLOAD != 'ingress-perf'){
```
Amoghrd, line 1108: same here

Jenkinsfile
Amoghrd: oh now I see L702 which checks for workload != None. can ignore this comment then

memodi: We check here https://github.com/memodi/ocp-qe-perfscale-ci/blob/0e8d0d6358cf66ad84b2711f59f767189ede8eef/Jenkinsfile#L701 prior before this condition is hit for workload is none

Jenkinsfile
memodi: I thought about it, but since we're not running ingress-perf as standalone for cron trigger, I thought it could just default to 3 replicas and there's always option to modify through param, wdyt?

PR #664, opened 2024-11-14
sync with prow ci

liqcui:
sync default instance type with prow ci

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -178,12 +178,12 @@ OPENSHIFT_ALERTMANAGER_STORAGE_SIZE=20Gi'''
               if(env.VARIABLES_LOCATION.indexOf("aws") != -1){
                 if (architecture_type.contains("arm64")) {
                       ENV_VARS += '''
-OPENSHIFT_INFRA_NODE_INSTANCE_TYPE=m6g.12xlarge
-OPENSHIFT_WORKLOAD_NODE_INSTANCE_TYPE=m6g.8xlarge'''
+OPENSHIFT_INFRA_NODE_INSTANCE_TYPE=m6g.4xlarge
+OPENSHIFT_WORKLOAD_NODE_INSTANCE_TYPE=m6g.4xlarge'''
               } else {
                   ENV_VARS += '''
-OPENSHIFT_INFRA_NODE_INSTANCE_TYPE=m5.12xlarge
-OPENSHIFT_WORKLOAD_NODE_INSTANCE_TYPE=m5.8xlarge'''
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -266,7 +280,7 @@ OPENSHIFT_WORKLOAD_NODE_INSTANCE_TYPE=n1-standard-32''')
                       string(name: 'DITTYBOPPER_REPO', value: DITTYBOPPER_REPO),
                       string(name: 'DITTYBOPPER_REPO_BRANCH', value: DITTYBOPPER_REPO_BRANCH),
                       string(name: 'DITTYBOPPER_PARAMS', value: DITTYBOPPER_PARAMS),
-                      text(name: 'ENV_VARS', value: ENV_VARS + '''
+                      text(name: 'ENV_VARS', value: ENV_VARS + '''                    
 OPENSHIFT_INFRA_NODE_VOLUME_SIZE=120
 OPENSHIFT_INFRA_NODE_CPU_COUNT=48
```
qiliRedHat, line 285: @liqcui I think this config is old and we did not update them recently. According to other platforms the below configs should be fine. Please also update the Prow default.
```
OPENSHIFT_INFRA_NODE_CPU_COUNT: "16"
OPENSHIFT_INFRA_NODE_MEMORY_SIZE: "65536"
```

liqcui: @qiliRedHat I noticed the workload with large instance size, like as 
PENSHIFT_WORKLOAD_NODE_VM_SIZE=n1-standard-32
Should we set a small size instance for each provider?

qiliRedHat: @liqcui Our tests don't need workload node anymore, there is no reference of workload node size now. Actually we can delete the workload related envs as we don't install workloads now.

liqcui: @qiliRedHat Let's set it to as minimal size and set it to false as default instead of deleting it, my ovn live migration still use workload node, just in case we can use it in some case in the future.

qiliRedHat: @liqcui Thanks for the comment. Agree, you can set the default size to what the ovn live migration test requires, that's the only test still needs it afaik.

liqcui: @qiliRedHat Update code and set workload to false and using small instance as default, please review again, thanks!
https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/liqcui-e2e-benchmarking-multibranch-pipeline/job/cluster-workers-scaling/957/console

PR #665, opened 2024-11-18
bug fix to update NOO_BUNDLE_VERSION

memodi:
Last PR introduced a bug where NOO_BUNDLE_VERSION is not captured properly, this PR fixes both cases when Installation method is None or otherwise, NOO_BUNDLE_VERSION will be left null or updated.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,10 @@ if (userId) {
 
 pipeline {
     environment {
-        def NOO_BUNDLE_VERSION = ''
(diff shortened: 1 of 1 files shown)


memodi:
tested here: https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/scale-ci/job/memodi-e2e-benchmarking-multibranch-pipeline/job/noo-bundle-vars/37/

PR #666, opened 2024-11-26
Do not count traffic returned from services in metrics + filter eth0

jotak:
- Traffic returned from service is duplicated with returned traffic from pods
- Filter on eth0 to avoid counting duplicates, cf comments in https://issues.redhat.com//browse/NETOBSERV-1954

### DIFF
--- a/scripts/queries/netobserv_prometheus_queries.yaml
+++ b/scripts/queries/netobserv_prometheus_queries.yaml
@@ -23,7 +23,7 @@
   metricName: nWorkloadBytesProcessedPerMinuteNetobserv
 
 # total bytes processed from workload namespaces by cadvisor / other means
-- query: sum(rate(container_network_receive_bytes_total{job="kubelet", metrics_path="/metrics/cadvisor", cluster="", namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*"}[1m])*60)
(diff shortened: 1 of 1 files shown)


jotak:
/cc @memodi 
/cc @Amoghrd 
The metric should not count Service traffic, which is duplicated with Pod traffic. We already excluded traffic from pods to services (e.g. an http request) but not from services to pods (e.g. an http response)

jpinsonneau (approved):

scripts/queries/netobserv_prometheus_queries.yaml
```diff
@@ -19,11 +19,11 @@
   metricName: nWorkloadFlowsProcessedPerMinuteTotals
 
 # total bytes processed from workload namespaces by netobserv
-- query: sum(rate(netobserv_workload_ingress_bytes_total{DstK8S_Namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*",DstK8S_Type!="Service"}[1m])*60)
+- query: sum(rate(netobserv_workload_ingress_bytes_total{DstK8S_Namespace=~"node-density-heavy.*|ingress-perf|cluster-density.*",SrcK8S_Type!="Service",DstK8S_Type!="Service"}[1m])*60)
```
memodi, line 22: @jotak - to make sure I understand this query, we're only looking to capture traffic that is pod to pod? 
what about traffic that's returned to the request from outside of cluster using openshift routes?

jotak: it's not only pod-to-pod, it's just not anything-to/from-service
So it also includes traffic from an external workload to internal pod.
Since we look at ingress with a destination namespace, it doesn't include the returned traffic, only the requests/incoming.

PR #670, opened 2024-12-04
Add bandwidth metrics to KPIs

jotak:
Adding two metrics to KPIs:
- `nWorkloadBytesProcessedPerMinuteNetobserv`, which tells bandwidth captured by netobserv
- `nWorkloadBytesProcessedPerMinuteCadvisor`, which tells bandwidth captured via container metrics / cAdvisor

Both of them are interesting, as they can tell:
- That netobserv captured the expected amount of traffic, ie. not dropping too much, and checking correctness at the same time, ie. we haven't introduced a regression that miscounts bytes
- That the workload traffic isn't affected by netobserv (ie. a drop in cAdvisor metrics is an indication of performance regression with latency induced by netobserv)

Those metrics are more reliable than the number of flows for these goals because the workload bandwidth is supposed to be somewhat consistant across runs, whereas number of flows is a more arbitrary measure that is only loosely related to the actual measured bandwidth.

### DIFF
--- a/scripts/queries/netobserv_touchstone_tolerancy_config.json
+++ b/scripts/queries/netobserv_touchstone_tolerancy_config.json
@@ -66,6 +66,32 @@
             },
             {
                 "filter": {
+                    "metric_name.keyword": "nWorkloadBytesProcessedPerMinuteNetobserv"
+                },
+                "buckets": [
+                    "metric_name.keyword"
+                ],
+                "aggregations": {
+                    "value": [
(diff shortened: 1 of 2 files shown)


jotak:
/cc @memodi 
/cc @Amoghrd 
Hey, here's a suggestion to add KPIs, cf explanation above

memodi (approved):
/lgtm
thanks @jotak

PR #682, opened 2025-01-13
update jenkins URL and loki cleanup fix

memodi:
update jenkins URL and loki cleanup fix

### DIFF
--- a/scripts/nope.py
+++ b/scripts/nope.py
@@ -42,9 +42,7 @@ END_TIME = None
 STEP = None
 
 # jenkins env constants
-JENKINS_URL = (
-    "https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/"
-)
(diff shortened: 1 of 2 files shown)


scripts/nope.py
```diff
@@ -801,6 +799,7 @@ def main():
         )
         try:
             JENKINS_SERVER = jenkins.Jenkins(JENKINS_URL)
+            version = JENKINS_SERVER.get_version()
```
Amoghrd, line 802: why is this field needed? I dont see it being used anywhere else

memodi: previous, L801 only created server object and didn't actually connect to jenkins, now with this line, it will actually try to connect to jenkins server and throw an exception if there's an error connecting to jenkins server. Due to this bug, we didn't notice any issue even though old URL no longer works and we have had our perf runs successfully passing this stage, however it didn't use to get workload information since it still kept using old jenkins URL and ES uploads was having incorrect workload name info.

Amoghrd: oh okay

PR #689, opened 2025-01-31
update for konflux builds

memodi:
- update pipelines and build_info script for konflux builds
- Look only severity=low agent errors


scripts/netobserv.sh
```diff
@@ -207,14 +207,12 @@ deploy_lokistack() {
 
 deploy_downstream_catalogsource() {
   echo "====> Creating brew-registry ImageContentSourcePolicy"
```
Amoghrd, line 209: Need to update this comment as well

PR #699, opened 2025-03-07
Adding arch_type

shahsahil264:
Adding arch type to avoid running into this error, when running multi-arch loaded upgrades 

```Could not find target version in 'https://amd64.ocp.releases.ci.openshift.org/graph' ```

more details here: https://github.com/openshift/release/pull/59743#issuecomment-2702905838

### DIFF
--- a/upgrade_scripts/upgrade.sh
+++ b/upgrade_scripts/upgrade.sh
@@ -16,7 +16,6 @@ scale=false
 eus=false
 eus_channel="fast"
 maxUnavail=1
-arch_type=""
 #optional parameters
 while [[ $# -gt 1 ]]
 do
@@ -49,11 +48,6 @@ case $key in
     shift # past argument
     shift # past value
     ;;
-    -a|--arch)
-    arch_type=$2
-    shift # past argument
-    shift # past value
-    ;;
     *)    # unknown option
     #need to get past file
     shift # past arg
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by shahsahil264** [comment]
Thanks for the comment Paige! 
I created a multi-arch cluster and checked the node_arc, its amd64. Hence, it'll never select the multi option in the script. 

```
oc version
Client Version: 4.14.1
Kustomize Version: v5.0.1
Server Version: 4.17.0-0.nightly-multi-2025-03-03-112159
Kubernetes Version: v1.30.10
➜  Documents oc get node | grep master| head -1| awk '{print $1}'
sahshah-multi-r4nv6-master-0
➜  Documents oc get node sahshah-multi-r4nv6-master-0 -ojsonpath='{.status.nodeInfo.architecture}'
amd64%. 
```

paigerube14:
> Thanks for the comment @paigerube14 I created a multi-arch cluster and checked the node_arc, its amd64. Hence, it'll never select the multi option in the script.
> ```

Let's add an extra check then and update [this if](https://github.com/openshift-eng/ocp-qe-perfscale-ci/blob/6dce199645124063208ab28efbf7d8f3c91f7925/upgrade_scripts/upgrade.sh#L84) statement instead of adding in another option. I think if we can find a way to see if its multi without passing a variable that'll be more useful and not hit issues in the future because its self sufficient.

shahsahil264:
> > Thanks for the comment @paigerube14 I created a multi-arch cluster and checked the node_arc, its amd64. Hence, it'll never select the multi option in the script.
> 
> Let's add an extra check then and update [this if](https://github.com/openshift-eng/ocp-qe-perfscale-ci/blob/6dce199645124063208ab28efbf7d8f3c91f7925/upgrade_scripts/upgrade.sh#L84) statement instead of adding in another option. I think if we can find a way to see if its multi without passing a variable that'll be more useful and not hit issues in the future because its self sufficient.

Sure on it!

paigerube14 (approved):
/lgtm

PR #703, opened 2025-03-14
Update flexy URL

Amoghrd:


memodi (approved):
/lgtm

Amoghrd:
Oh it seems like only @mffiedler and @qiliRedHat are approvers for this branch

paigerube14:
> Oh it seems like only @mffiedler and @qiliRedHat are approvers for this branch

@skordas it might be good if you could add yourself as an approver to each branch

PR #707, opened 2025-03-27
Not using the highest version on numpy

skordas:
For some time all [jobs](https://jenkins-csb-openshift-qe-mastern.dno.corp.redhat.com/job/scale-ci/job/e2e-benchmarking-multibranch-pipeline/job/post-results-to-es/) were failing because numpy version jumped over ES requirements.
We can try to using newer version of ES - but this should work.


skordas:
All jobs finishing with error: 
AttributeError: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.

PR #709, opened 2025-03-31
use ISO timestamps and set warning loglevel for es

memodi:
use ISO timestamps and set warning loglevel for es

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1110,6 +1110,12 @@ def setTemplateParams(){
     return templateParams
 }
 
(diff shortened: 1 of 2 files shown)


memodi:
Did a test run where timestamps conversions are performed well and also fixed bug to read baseline.json from new path:
https://jenkins-csb-openshift-qe-mastern.dno.corp.redhat.com/job/scale-ci/job/memodi-e2e-benchmarking-multibranch-pipeline/job/fix-timestamp/3/

PR #727, opened 2025-05-20
fix build_info.sh

memodi:
fix build_info.sh


Amoghrd:
The same fix needs to be committed in netobserv.sh file of release and netobserv-qe-scripts repo too right?
/lgtm

Amoghrd:
I was talking about netobserv.sh file changes for Downstream var in  [netobserv-qe-scripts](https://gitlab.cee.redhat.com/netobserv-qe/netobserv-qe-scripts/-/blob/main/netobserv.sh?ref_type=heads#L240) and in [release](https://github.com/openshift/release/blob/master/ci-operator/step-registry/netobserv/install-operator/netobserv-install-operator-commands.sh#L123) repos

PR #760, opened 2025-06-06
Netobserv: track metrics cardinality

jotak:
cc @memodi @Amoghrd


scripts/queries/netobserv_touchstone_tolerancy_rules.yaml
```diff
@@ -61,6 +61,9 @@
 - json_path: ["metric_name", "rssPrometheusTotals", "metric_name", "*", "avg(value)"]
   tolerancy: 60
   max_failures: 0
+- json_path: ["metric_name", "metricsCardinality", "metric_name", "*", "avg(value)"]
+  tolerancy: 10
```
memodi, line 65:
```suggestion
  tolerancy: 30
```

recommend some higher threshold as we probably don't want entire comparison to fail on this metric.

memodi (changes requested):

PR #764, opened 2025-06-13
replace oc wait with while loop for ebpf pods check

memodi:
replace oc wait with while loop for ebpf pods check since pods name change if they restart while `oc wait` causes it to fail.


scripts/env.sh
```diff
@@ -1,5 +1,5 @@
 # use upstream version for perf and reliability testing
```
Amoghrd, line 1: can update the comment here and also the comment for TOPIC_PARTITIONS

PR #799, opened 2025-07-16
Reapply "Netobserv: track metrics cardinality (#760)" (#765)

jotak:
This reverts commit 802349789c639347816da2ce47c07fc7c4cbe4cc.

@memodi , I'd like to revisit adding metrics cardinality as this is something important to know when tracking down potential regressions. I will bring some changes that I hope will fix the issue that you saw - and if it doesn't, we should try to investigate/refine on live cluster, if possible, rather than reverting?


jotak:
> could you add these metrics to new orion config files as well?

rebased & done
For info I've set direction to 0: both increased or decreased cardinality is a smell of something wrong if not intentional.

PR #805, opened 2025-08-06
Remove ping capability

mcornea:
By default ping is shipped with cap_net_raw=ep capability which prevents ping from running in unprivileged pods which results in CI errors such as:

```
+ ping -c 5 XXXXXXXXXX
/bin/bash: line 11: /bin/ping: Operation not permitted
```

This change removes the ping binary capability which should allow running the ping checks in the CI steps.


paigerube14 (approved):
/lgtm

PR #806, opened 2025-08-07
NETOBSERV-2353: KafkaNodePool configs

Amoghrd:
Add KafkaNodePool configs for NetObserv


Amoghrd:
Based on testing PR https://github.com/openshift/openshift-tests-private/pull/26515 the configs are working as expected. @mffiedler Could you PTAL?

PR #811, opened 2025-10-07
Update dashboard

Amoghrd:


scripts/queries/netobserv_prometheus_queries.yaml
```diff
@@ -218,11 +218,11 @@
 # Total metrics #
 ###########################
 
-# Total NetObserv CPU Usage, assuming Loki and Kafka also runs in netobserv NS
```
memodi, line 221: suggest to update: 
"assuming Loki and Kafka also runs within NS with 'netobserv' name in them"

